PR #14, opened 2022-03-11
Support multiple fields per Item in a Capture

### LINKED ISSUE #7: Switch to ColumnView in GTK frontend, to display multiple fields for each Item

miek:
(Marked draft for now as I'll probably need to rebase this after #12 is merged)

This PR adds a method `get_fields` on `Capture` which can return multiple fields as well as the current summary, so that we can display timestamps, ACK/NAK status, etc.

I wanted to set this up in a way that doesn't require the frontend to know all the field names, as it would be cumbersome to update as we add fields. So, this also includes a [procedural macro](https://doc.rust-lang.org/reference/procedural-macros.html) that lets us annotate each field with its name (for column headers in the UI) and whether it should display the TreeExpander widget. The macro then adds some methods that the frontend can use to iterate over the available fields.

fixes #7

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -104,6 +104,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
+name = "derive_macro"
+version = "0.1.0"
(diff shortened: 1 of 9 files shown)


martinling:
I don't know why this was just closed. I didn't knowingly do anything to close it. But it's long stale at this point anyway.

PR #16, opened 2022-03-23
Remove finish method

martinling:
The only work still needed in the `finish` method was to add an entry to the endpoint’s transfer index, if a transfer was left in progress at the end of capture.

This step is moved to `transfer_start`, the `finish` method is removed, and `ep_transfer_end` is merged into `transfer_end`.


miek:
Thanks!

PR #17, opened 2022-03-24
Replace use of TreeExpander

martinling:
This turned out to be fairly simple. We can just replace the `TreeExpander` with a generic `Expander`, show or hide it according to  `TreeListRow::is_expandable`, and connect its expansion to `TreeListRow::set_expanded`.

This fixes the duplicate indent that was being added by `TreeExpander`.

I've adjusted the connecting lines so that they line up with the expander widgets. It's not a perfect match, but I guess it'll do until we have a graphical version of the connecting lines.

### DIFF
--- a/src/capture.rs
+++ b/src/capture.rs
@@ -788,11 +788,11 @@ impl Capture {
         }
         connectors.push_str(
             match (item, last_packet) {
-                (Transfer(_), _) if entry.is_start() => "─▷",
-                (Transfer(_), _)                     => "─□",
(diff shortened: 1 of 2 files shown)


martinling:
This also ties in with keeping track of which items are expanded in the view, and thus knowing what may need to be updated during live capture.

Since it's now our handler connecting the expander widget with the model, we can maintain a set of what is currently expanded, and add or remove items from the set in the handler.

PR #27, opened 2022-05-05
Transaction grouping

martinling:
This stacks on #26 and adds an extra layer of the hierarchy which can be used to group transactions within a transfer.

This allows NAKed transactions to be grouped together, making it easier to see actual traffic:

image


martinling:
Closing in favour of the new approach in #37.

PR #29, opened 2022-05-10
Add tests using reference captures

martinling:
This PR adds tests based on reference captures.

Each test is a subdirectory under `tests` containing a `capture.pcap` and `reference.txt` file.

The `reference.txt` file contains the expected summary text for every item in the capture, indented by one space for each level in the hierarchy. The test harness processes `capture.pcap`, creates an `output.txt` file in the same directory, and then compares the two text files line by line.

To add a new test, create a subdirectory with a `capture.pcap` file and an empty `reference.txt` file, then run `cargo test`. The test will fail, but the `output.txt` file is created and can then be copied over `reference.txt`.

To update a test after output has been changed, copy `output.txt` over `reference.txt`.


martinling:
Note that the `hackrf-connect` test contains some unexpected results: the final transaction in the last transfer is shown as including SOF packets that follow the ACK. I'm not sure why this happens, but it doesn't seem to be a recently introduced bug. I suggest we get the test infrastructure merged before trying to fix it, and update the expected output once it's fixed.

PR #31, opened 2022-05-17
Setup GitHub Actions

miek:
This sets up GitHub actions to build & test any pushes or PRs. An example run can be seen here: https://github.com/miek/packetry/actions/runs/2339638314

It also adds a [Clippy](https://github.com/rust-lang/rust-clippy#clippy) pass, which has a bunch of lints to check for style etc. I think the suggestions are pretty good (e.g: https://github.com/miek/packetry/runs/6453987836?check_suite_focus=true), though I did squelch a few where I think it made more sense to break the rules for code clarity or to match USB spec naming.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,76 @@
+on:
+  push:
+  pull_request:
+
+name: CI
+
+jobs:
+  clippy:
+    name: Clippy
+    runs-on: ubuntu-22.04
+
+    steps:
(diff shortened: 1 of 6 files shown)


martinling:
It looks like this is only running `cargo build` and not `cargo test`, unless I'm missing something?

miek:
Doh! Yep, you're right - I thought the config I was borrowing from had it in there and hadn't noticed. Fixed!

PR #33, opened 2022-05-20
Use specialised types to represent indices instead of bare integers

martinling:
This PR adds new numeric types, to replace the use of bare integer types as indices into arrays and streams. This improves the readability of the code and allows the compiler to catch misuse of indices.

The `u64` Indices used by `FileVec` and `HybridIndex` are replaced with a new `Id<T>` type associated with the type of data being indexed. Type aliases for our uses of `Id<T>` are added for clarity, e.g:
- `DeviceId` is a type alias for `Id<Device>`.
- `PacketId` is a type alias for `Id<Id<u8>>>`, because the packet index contains offsets into the packet data stream.
- `EndpointTransferId` is a type alias for `Id<Id<Id<Id<Id<u8>>>>>` (!), because there is an index chain through the endpoint transfer index, endpoint transaction index, transaction index and packet index, ending in an offset into the packet data stream.

The `u8` indices used in the `usb` module are replaced with newtypes such as `DeviceAddr` and `EndpointNum` which are tuple structs containing a `u8` value. The [derive_more](https://crates.io/crates/derive_more) crate is used to simplify implementing `From<u8>` and other required traits on these wrappers.


martinling:
Rebased this to make it easier to see what's changed, now that #32 is merged.

PR #34, opened 2022-05-30
Further improvements to types and Rust usage in capture & decoder modules

martinling:
This PR builds on the work in #33. The common theme is to make better use of the Rust type system in order to streamline the code in the `capture` & `decoder` modules and, more generally, to write better Rust.

The main changes include:

- Add further types to the `usb` module to encapsulate more conversion and formatting details. These include `EndpointAddr`, `EndpointType`, `StringId`, `BCDVersion`, `UTF16Bytes` and `UTF16ByteVec`.

- Add a `VecMap<K, V>` type to use in place of `Vec`. This has two key properties:
  - Indices can be of a specific newtype, converted to `usize` internally. This removes a lot of `x as usize` conversions in the code.
  - Items can be inserted at a specific index, with the underlying `Vec<Option<V>>` being enlarged as necessary. This removes several places where we were doing this by hand.

- Prevent the decoder from peeking ahead at what the ID of a stream item will be. Instead, return an ID only when an item is pushed out to the stream. This enforces the property that no item can be referred to by ID from another stream, before it has been output itself. Several parts of the decoder were not keeping to this rule and have been refactored accordingly.

- Rename methods in line with Rust API guidelines, in particular eliminating a lot of `get_` prefixes.

- Add preludes to the `usb` and `capture` modules to avoid lengthy import lists.

### DIFF
--- a/src/capture.rs
+++ b/src/capture.rs
@@ -125,6 +125,9 @@ pub enum EndpointState {
     Ending = 3,
 }
 
+pub const INVALID_EP_NUM: u8 = 0x10;
+pub const FRAMING_EP_NUM: u8 = 0x11;
+
 #[derive(Copy, Clone, Debug)]
 pub enum EndpointType {
     Unidentified,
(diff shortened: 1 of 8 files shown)


martinling:
Rebased to fix conflicts.

miek (approved):

PR #36, opened 2022-06-15
Integrated version of tree model work

martinling:
This intergrates @miek's `tree-model` branch with the previous generic UI code, allowing the `TreeListModel` and associated UI code to be shared between the traffic view and the device hierarchy view.


miek:
Merged in 48b838cbb1d36ca4b383518074f4e54a5fc565b4

PR #38, opened 2022-09-14
Clippy & CI fixes

miek:
This fixes a handful of Clippy warnings that were introduced by a new Rust version.

For most of them, I've implemented the suggested fix.
For the [type_complexity](https://rust-lang.github.io/rust-clippy/master/index.html#type_complexity) warning, I tried to break out part of it as a new `type` but ran into a problem where it needed an unstable feature for inherent associated types, so I've just squelched that warning.

The other commit sets the workflow to run every week as we have on most of the other repos, so we can spot issues like this sooner. It should also allow for the workflow to be triggered manually.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,6 +1,11 @@
 on:
   push:
   pull_request:
+  workflow_dispatch:
+
+  # Run automatically every monday
+  schedule:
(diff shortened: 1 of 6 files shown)


martinling (approved):
Looks good, thanks for chasing this up.

PR #39, opened 2022-09-20
Add initial live capture support & LUNA backend

miek:
This adds support for live capture and adds a basic LUNA backend using rusb. When Packetry is launched without a filename argument, it will attempt to connect to a LUNA device running the USB analyzer applet and begin capture.

There are a few outstanding issues, but I think it's better to get this merged then fix those in-tree later:
 * We discussed using the new `item_end` method to check whether an Item is complete, but I had some trouble making it work as I don't have the right type available for the lookup (I have an `Item` not an `ItemId`). I noticed the method signature has changed in the `interleaving` branch, so I left this as a todo for after that merge.
 * It doesn't currently handle updates other than at the top-level of the tree.
 * There's no logic yet to query whether to update a DeviceItem, so it is always queried for updates.


src/backend/luna.rs
```diff
@@ -0,0 +1,100 @@
...
+            LunaDevice::enable_capture(&handle);
+            loop {
+                let result = handle.read_bulk(ENDPOINT, &mut buffer, Duration::from_secs(1));
+                if let Ok(count) = result {
+                    packet_queue.extend(&buffer[..count]);
+                    while let Some(packet) = packet_queue.next() {
+                        tx.send(packet).expect("Failed to send packet across channel");
+                    }
+                }
+            }
```
martinling, line 49: This should do some amount of error handling. As it stands you can unplug the analyzer and it will just peg the CPU spinning in this loop. We should continue when `result` is `Err(Timeout)`, but any other error ought to cause this loop to exit.

src/backend/luna.rs
```diff
@@ -0,0 +1,100 @@
...
+
+pub struct LunaDevice {
+    handle: Option<DeviceHandle<GlobalContext>>,
+    worker: Option<thread::JoinHandle<DeviceHandle<GlobalContext>>>,
+    receiver: Option<Receiver<Vec<u8>>>,
+}
+
+impl LunaDevice {
+    pub fn open() -> Self {
+        let handle = rusb::open_device_with_vid_pid(VID, PID).unwrap();
```
martinling, line 24: I'd rather not see `unwrap` inside the implementation here - let's define an error type and pass it back to `main`.

src/tree_list_model.rs
```diff
@@ -138,62 +150,76 @@ where Item: Copy,
...
+        } else {
+            model.items_changed(position, node.total_child_count, 0);
         }
 
-        node_ref.borrow_mut().expanded = expanded;
         Ok(())
     }
 
+    pub fn update(&mut self) -> Option<(u32, u32, u32)> {
+        let mut cap = self.capture.lock().unwrap();
```
martinling, line 196: I realise this gets called from an the Glib idle handler which is an awkward place to handle errors, but let's try to keep things as clean as we can - we can still return a `Result` here, and put the `unwrap` in the idle handler.

src/backend/luna.rs
```diff
@@ -0,0 +1,100 @@
...
+        LunaDevice {
+            handle: Some(handle),
+            worker: None,
+            receiver: None,
+        }
+    }
+
+    pub fn start(&mut self) {
+        let mut handle = self.handle.take().unwrap();
```
martinling, line 33: Again, let's return a `Result` here.

src/backend/luna.rs
```diff
@@ -0,0 +1,100 @@
...
+    fn enable_capture(handle: &DeviceHandle<GlobalContext>) {
+        use rusb::{Direction, RequestType, Recipient, request_type};
+        handle.write_control(
+            request_type(Direction::Out, RequestType::Vendor, Recipient::Device),
+            1,
+            1,
+            0,
+            &[],
+            Duration::from_secs(5),
+        ).expect("write_control failed in enable_capture");
```
martinling, line 62: Similarly, this ought to return a `Result` rather than using `expect`.

src/backend/luna.rs
```diff
@@ -0,0 +1,100 @@
...
+
+const ENDPOINT: u8 = 0x81;
+
+const READ_LEN: usize = 0x4000;
+
+pub struct LunaDevice {
+    handle: Option<DeviceHandle<GlobalContext>>,
+    worker: Option<thread::JoinHandle<DeviceHandle<GlobalContext>>>,
+    receiver: Option<Receiver<Vec<u8>>>,
+}
```
martinling, line 20: Rather than having all these be `Option`, and potentially `None` if you have the thing in the wrong state, could we split this structure up?

I'm thinking that `start()` could return a `LunaCapture` which would own the worker thread and the receiver, and then `next()` could be on the `LunaCapture`, which would avoid the need to do `self.receiver.as_ref().unwrap()` in there.

martinling (changes requested):
Great work making this happen! Sorry it's taken me a while to get to reviewing it with all the HackRF stuff going on.

This is working okay for me, though capture does fall over quite quickly with any significant amount of traffic. That's to be expected though given the current small BRAM buffer and the use of the synchronous `rusb` API, which I think is the right thing to do for now for an initial version. We can see about using the experimental async stuff later.

I've commented on a few things I'd like to see tweaked before merge. Mostly just suggested cleanups; the only one I'm particularly concerned with is items not getting removed from `incomplete_items`.

src/model/mod.rs
```diff
@@ -20,6 +21,17 @@ glib::wrapper! {
     pub struct DeviceModel(ObjectSubclass<imp::DeviceModel>) @implements gio::ListModel;
 }
 
+impl TrafficModel {
+    pub fn update(&self) {
+        let mut tree_opt  = self.imp().tree.borrow_mut();
+        let tree = tree_opt.as_mut().unwrap();
+        if let Some((position, _, added)) = tree.update().unwrap() {
```
martinling, line 28: The first `unwrap` here is fine, since that `Option` is just there for annoying GObject reasons and we know it can't actually be `None`.

I think the second one should still be moved one step further out, into the idle handle closure in `main.rs`.

martinling (approved):

martinling (changes requested):
Actually, having tested a bit more I'm finding the analyzer isn't always shut down on normal exit. I'm working on a fix.

martinling:
Proposed fix here: https://github.com/martinling/packetry/commit/d58ca5c806de6935a5c25bc210e6ecee380e50c8

miek:
> Proposed fix here: [martinling@d58ca5c](https://github.com/martinling/packetry/commit/d58ca5c806de6935a5c25bc210e6ecee380e50c8)

Looks good to me.

martinling (approved):
Let's get this merged then!

PR #41, opened 2022-10-26
Improve error reporting

martinling:
Two changes here:

1. We use `IndexError` as a catch-all for any lookup failure in `Capture`. Make it take a string argument to indicate the cause, and add suitable strings at all call sites.

2. When lookup for an individual tree node fails, display the error in the view rather than crashing.


miek:
Looks good to me, thanks!

I cherry-picked the clippy fix already so it stops breaking other PRs. Could you rebase this and I'll merge it in.

martinling:
Rebased and should be ready to merge now.

PR #43, opened 2022-11-02
More comprehensive error handling

martinling:
This PR aims to ensure that everywhere an error can occur, it results in a dialog box with a useful message, rather than a panic or printing a message on stdout.

There are a few `expect` calls left in places where we call `glib::Object::new`. These will go away when we update to [gtk-rs 0.5](https://gtk-rs.org/blog/2022/10/18/new-release.html), which changes the return type of that function to no longer return a `Result`.

I haven't included that update as part of this PR because it depends on rustc 1.63 which is still quite new.

 Best reviewed commit-by-commit. Ways to see errors for testing:
- Give a filename that doesn't exist or isn't valid pcap.
- Start without a filename and no LUNA connected.
- Disconnect LUNA during capture.


martinling:
Rebased on main.

martinling:
Added two more commits:

1. Don't automatically close the main window after displaying an error dialog. In most cases the application is still otherwise usable.

2. Don't bail out of the UI setup on a pcap or decoder failure, just display an error and break out of the read loop at that point. This allows the application to proceed with displaying as much of the capture as could be read and decoded successfully.

PR #47, opened 2022-11-13
Handle PING transactions in decoder

martinling:
This adds handling of PING transactions to the transaction and transfer state machines.

I've added a test case of enumerating a HackRF in DFU mode, which NAKs some OUT transactions during control transfers, leading the host to need to PING once before retrying.


src/decoder.rs
```diff
@@ -107,6 +107,9 @@ impl TransactionState {
             // OUT may also be completed by NAK or STALL.
             (Some(OUT), Some(DATA0 | DATA1), NAK | STALL) => Done,
 
+            // PING may be followed by ACK or NAK.
+            (Some(PING), Some(PING), ACK | NAK) => Done,
+
```
miek, line 112: Should this also match on `STALL`? (based on Fig 8-28 / Fig 8-30 in the spec)

martinling: It looks like it should, yeah.

I find figure 8-28 and the related state machine diagrams quite confusing; they seem more like they're illustrating implementation details of a reference implementation rather than the underlying protocol logic.

In the text of 8.5.1 it only says *"The endpoint either responds to the PING with a NAK or an ACK handshake"*.

But figure 8-30 is pretty clear that PING -> STALL is allowed, so I've added it.

src/decoder.rs
```diff
@@ -81,43 +82,48 @@ impl TransactionState {
...
         use PID::*;
         // A transaction is completed if it has 3 valid packets and is
-        // acknowledged with an ACK or NYET handshake.
-        match (self.count, self.last) {
-            (3, Some(ACK | NYET)) => true,
-            (..)                  => false
+        // acknowledged with an ACK or NYET handshake, except for PING
+        // transactions which complete after 2 with an ACK or NAK.
+        match (self.first, self.count, self.last) {
+            (Some(PING), 2, Some(ACK | NAK) ) => true,
```
miek, line 124: I guess this should match `STALL` too?

martinling: No, I don't think it should. This method is used in `transfer_status` to decide whether the transfer should progress to the next state. For other transactions that happens on an ACK or NYET, which both indicate success. For PING, I accepted both ACK and NAK because either way the ping is complete, but actually I don't think that's correct; if the PING response is a NAK then the host is supposed to PING again before it can proceed to send OUT. So actually I think we should only accept ACK here.

PR #49, opened 2022-11-24
TreeListModel improvements

martinling:
This PR stacks on #50 and collects some improvements to the existing TreeListModel code. The key changes are:

1. Define separate types for the root node and item nodes, with a common `Node` trait, and a `Children` structure used by all three. This allows the same code to be used to access the children of either, and avoids needing to handle the root node case in contexts where we know we are dealing with an item node.
2. When the user expands or collapses an item, pass its absolute position, which we can retrieve from the `ListItem`, into `set_expanded`. This saves us needing to recalculate it from the tree and removes the need for the `relative_position` method.
3. Return an error if `set_expanded` is called to expand a node that is already expanded, or vice versa.
4. Decouple the `TreeListModel` implementation from its GTK wrapper:
    1. Rather than having `set_expanded` call into the `items_changed` method of the `ListModel`, have it return a `ModelUpdate` struct giving the row count changes, then handle the GTK-specific details in the wrapper. This allows us to remove the `Model` type parameter from the definition of `TreeListModel`. The live `update` method is also modified to return a `ModelUpdate`.
    2. Remove the `n_items` and `item` methods from `TreeListModel`, and have the wrapper call directly into `fetch` and a new `row_count` method. This allows us to remove the `RowData` type parameter also, making the `TreeListModel` implementation fully independent of the GTK API.
    3. Use `u64` indices in `TreeListModel`, putting the necessary adjustments for compatibility with GTK's `u32`  indices into the GTK wrapper instead.
5. Remove the `complete` methods, which are not currently used. These will be replaced with a new API in the next PR.

This PR is the first in a series of three which will eventually introduce the interleaved tree model. The second PR will reimplement `TreeListModel` with a region map, and ensure that all required UI updates occur during live capture. The third PR will introduce the interleaving.

### DIFF
--- a/src/capture.rs
+++ b/src/capture.rs
@@ -253,12 +253,6 @@ impl Transaction {
             (..)            => false
         }
     }
-
-    fn complete(&self) -> bool {
-        use PID::*;
-        // TODO: iso transactions end on a DATAx
-        matches!(self.end_pid, ACK | NAK | NYET | STALL | ERR)
-    }
 }
 
 pub fn fmt_count(count: u64) -> String {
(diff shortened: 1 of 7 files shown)


martinling:
Changing this one back to draft for now, because the work to decouple the tree model from the GTK details makes implementing live update of arbitrary sets of items a lot more complicated. I'm working on another branch that may simplify things.

martinling:
This PR is obsolete, I have a new set of branches to follow after #50 and #51 are merged.

PR #61, opened 2023-02-16
Lock-free stream implementation

martinling:
This PR adds a new lock-free implementation of the basic underlying data structure required by Packetry: an append-only data stream, backed by a file, for which we require fast random access to any part of the data whilst it continues to grow.

Packetry's current implementation, in the `FileVec` and `HybridIndex` types, uses file I/O which adds significant overhead. Both reading and writing require exclusive access to a stream object, which is currently managed by a single `Mutex` protecting the whole `Capture` database. This creates contention between the decoder and UI threads.

The new implementation creates two objects for each stream: a unique `StreamWriter`, and a clonable `StreamReader`. The writer and any number of readers may operate concurrently without blocking each other.

- `StreamReader` creates memory-mapped regions on demand to access the file in blocks of a configurable size, which must be a multiple of the system page size. Each reader maintains a fixed-size cache of mappings for its most recently used blocks, using the [LruBTreeMap](https://docs.rs/lrumap/0.1.0/lrumap/struct.LruBTreeMap.html) type.

- `StreamWriter` operates much like a conventional buffered writer, but uses interchangeable write buffers whose lifetimes are managed by `Arc`. To read stream data not yet written to the file, readers clone the `Arc` of the current write buffer, which they can then retain access to after the writer moves on to using another buffer.

Atomically switching from one write buffer to another is achieved using the [ArcSwap](https://docs.rs/arc-swap/latest/arc_swap/) type.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,5 +15,8 @@ pub mod ui;
 mod usb;
 mod vec_map;
 
+#[cfg(test)]
+mod stream;
+
 #[cfg(any(feature="test-ui-replay", feature="record-ui-test"))]
 pub mod record_ui;
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -8,7 +8,7 @@ use std::slice;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicU64, Ordering::{Acquire, Release}};
 
-use arc_swap::ArcSwap;
+use arc_swap::{ArcSwap, ArcSwapOption};
 use lrumap::LruBTreeMap;
 use memmap2::{Mmap, MmapOptions};
 use tempfile::tempfile;
@@ -22,7 +22,7 @@ struct Shared<const S: usize> {
     /// Available length of the stream, including data in both file and buffer.
     length: AtomicU64,
     /// File handle used by readers to create mappings.
-    file: File,
+    file: ArcSwapOption<File>,
     /// Buffer currently in use for newly appended data.
     current_buffer: ArcSwap<Buffer<S>>,
 }
@@ -38,7 +38,7 @@ pub struct StreamWriter<const S: usize = MIN_BLOCK> {
     /// Pointer to current position in the current buffer.
     ptr: *mut u8,
     /// File handle used to append to the stream.
-    file: File,
+    file: Option<File>,
     /// Spare buffer to be potentially used when current buffer is full.
     spare_buffer: Option<Arc<Buffer<S>>>,
 }
@@ -103,7 +103,7 @@ pub fn stream<const S: usize>() -> Result<StreamPair<S>, StreamError> {
     let buffer = Arc::new(Buffer::new(0)?);
     let shared = Arc::new(Shared {
         length: AtomicU64::from(0),
-        file: tempfile().map_err(StreamError::TempFile)?,
+        file: ArcSwapOption::empty(),
         current_buffer: ArcSwap::new(buffer.clone()),
     });
     let writer = StreamWriter {
@@ -111,7 +111,7 @@ pub fn stream<const S: usize>() -> Result<StreamPair<S>, StreamError> {
         length: 0,
         buf: buffer.ptr,
         ptr: buffer.ptr,
-        file: shared.file.try_clone().map_err(StreamError::CloneFile)?,
+        file: None,
         spare_buffer: None,
     };
     let reader = StreamReader {
@@ -196,8 +196,19 @@ impl<const BLOCK_SIZE: usize> StreamWriter<BLOCK_SIZE> {
     ///
     unsafe fn write_to_file(&mut self, data: &[u8]) -> Result<(), StreamError> {
 
+        // Create the file if it does not exist yet.
+        let file = match &mut self.file {
+            None => {
+                let file = tempfile().map_err(StreamError::TempFile)?;
+                self.shared.file.store(Some(Arc::new(
+                    file.try_clone().map_err(StreamError::CloneFile)?)));
+                self.file.insert(file)
+            },
+            Some(file) => file
+        };
+
         // Write the data to file.
-        self.file.write(data).map_err(StreamError::WriteFile)?;
+        file.write(data).map_err(StreamError::WriteFile)?;
 
         // We must change the stream's current buffer to one for the new block.
         let block_base = self.length;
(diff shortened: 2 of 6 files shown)


src/stream.rs
```diff
@@ -0,0 +1,517 @@
...
+
+    /// Layout for allocating block-sized, block-aligned chunks of memory.
+    const fn block_layout() -> Layout {
+        assert!(BLOCK_SIZE != 0);
+        assert!(BLOCK_SIZE.is_power_of_two());
+        assert!(BLOCK_SIZE < (isize::MAX as usize));
+        // Safety: Align is non-zero, a power of 2, and does not overflow isize.
+        unsafe {
+            Layout::from_size_align_unchecked(BLOCK_SIZE, BLOCK_SIZE)
+        }
```
miek, line 367: If we're willing to require Rust 1.57 (which stabilised const_panic), this unsafe block could be replaced with something like:
```rust
        match Layout::from_size_align(BLOCK_SIZE, BLOCK_SIZE) {
            Ok(x) => x,
            _ => panic!("..."),
        }
```
The only major distro I can see that's too old is current Debian stable (bullseye), but bookworm should be out very soon.
Otherwise, we could leave it as a TODO for the future.

martinling: With this PR merged we'll require at least 1.51 for const generics, which already exceeds the 1.48 in bullseye. I suspect that our MSRV may already be above that in practice. I've been building with 1.63 for quite a while, which is what's going to be in bookworm. I'm not sure we actually require it for anything yet, but I've not been checking that. How do the versions look on the other distros you looked at - what would be a sensible MSRV at this point?

> ```rust
> match Layout::from_size_align(BLOCK_SIZE, BLOCK_SIZE) {
>     Ok(x) => x,
>     _ => panic!("..."),
>    }
> ```

One issue with this approach in general is that the `panic` message can't display the error because `fmt` isn't `const`, so you can only give a generic message.

However, since there's a very limited set of possible in this case, we can do:

```rust
match Layout::from_size_align(BLOCK_SIZE, BLOCK_SIZE) {
    Ok(layout) => layout,
    Err(_) => panic!("{}",
        if BLOCK_SIZE == 0 {
            "Stream block size must not be zero"
        } else if !BLOCK_SIZE.is_power_of_two() {
            "Stream block size must be a power of two"
        } else if BLOCK_SIZE > (isize::MAX as usize) {
            "Stream block size must be within isize::MAX"
        } else {
            "Unexpected layout error"
        }
    )
}
```

miek: > With this PR merged we'll require at least 1.51 for const generics, which already exceeds the 1.48 in bullseye. I suspect that our MSRV may already be above that in practice. I've been building with 1.63 for quite a while, which is what's going to be in bookworm. I'm not sure we actually require it for anything yet, but I've not been checking that. How do the versions look on the other distros you looked at - what would be a sensible MSRV at this point?

Recent versions of Ubuntu/Fedora ship at least 1.65, I think the oldest one worth worrying about is 1.59 in openSUSE Leap 15.4. There's a good overall summary here: https://repology.org/project/rust/badges

Right now we still build successfully with 1.47 (https://github.com/miek/packetry/actions/runs/4447312275) so I think setting an MSRV of 1.57 for const_panic + const generics would be fine, and there's room to bump it up if we find something else we need.

martinling: Cool!

I've squashed the `from_size_align` change into the commit that adds the const generics.

I've added a commit on the end which documents our MSRV as 1.57, and adds 1.57 to the CI matrix. Conveniently, Rust 1.56 added a `rust-version` field to `Cargo.toml`, so it's formally specified there there as well.

martinling:
I've added a runtime check that the block size is a multiple of the system page size.

src/lib.rs
```diff
@@ -15,5 +15,8 @@ pub mod ui;
 mod usb;
 mod vec_map;
 
+#[cfg(test)]
```
antoinevg, line 18: I thought this was rather a nice use of `#[cfg(test)]` from both a convenience and semantic point of view.

src/stream.rs
```diff
@@ -0,0 +1,533 @@
...
+        // Update shared length value, and return new length.
+        self.shared.length.store(self.length, Release);
+        Ok(self.length)
+    }
+
+    /// Helper method for writing data to buffer.
+    ///
+    /// The data must be guaranteed to fit within the remaining space.
+    ///
```
antoinevg, line 185: I recently noticed a section in the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/documentation.html) book that recommends adding an explicit "Safety" heading for documenting `unsafe` usage.

e.g. this would just become:

```
    /// Helper method for writing data to buffer.
    ///
    /// # Safety
    ///
    /// The data must be guaranteed to fit within the remaining space.
    ///
```

martinling: Yes, it would probably be better to be more explicit here, both about the fact this is the safety condition, and also what code is responsible for ensuring it. I think we can do without the heading when the comment fits on one line, though:
```
/// Safety: The caller must ensure that the data will fit within the remaining space in the buffer.
```

src/stream.rs
```diff
@@ -0,0 +1,533 @@
...
+    unsafe fn write_buffer_to_file(&mut self) -> Result<(), StreamError> {
+        let buf = slice::from_raw_parts(self.buf, Self::block_size());
+        self.write_to_file(buf)
+    }
+
+    /// Helper method for writing data to file.
+    ///
+    /// The data must be guaranteed to be a multiple of the block size.
+    ///
+    unsafe fn write_to_file(&mut self, data: &[u8]) -> Result<(), StreamError> {
```
martinling, line 206: This function is a little annoying: it is `unsafe` to call because it requires a guarantee that the data length is a multiple of the block size, but the code within it is actually entirely safe.

I understand that Rust plans to eventually have `unsafe` functions not automatically mark their scope `unsafe`, but it's not done yet: https://github.com/rust-lang/rust/issues/71668

martinling: Oh, actually we can have that now with `#[deny(unsafe_op_in_unsafe_fn)]`.

src/stream.rs
antoinevg: Cosmetically I'd argue for using a heading so the rendered docs stay in line with any of the other "official" doc sections - i.e.

screenshot

antoinevg: Oh I take that back, clippy has some:

https://rust-lang.github.io/rust-clippy/rust-1.68.0/index.html#missing_errors_doc
https://rust-lang.github.io/rust-clippy/rust-1.68.0/index.html#missing_panics_doc
https://rust-lang.github.io/rust-clippy/rust-1.68.0/index.html#missing_safety_doc

martinling: The `unsafe` functions here aren't public, so shouldn't be showing up in any docs - they're just internal helpers used to implement the public `append` method.

src/lib.rs
martinling: It was also essential, or you get warnings on a regular build about the code being unused.

antoinevg (approved):
👍

PR #54, opened 2022-12-23
Interleaved traffic view

martinling:
This draft PR collects my work to make the traffic view fully interleaved, i.e. transactions and packets are shown in correct chronological order, when overlapping transfers are expanded in the view.

Making this work correctly and efficiently requires complex modifications to the tree model, and the implementation of several new queries to the capture database.

The commits on this branch are in several sub-branches, which will be merged in separate PRs:

1. [x] #51
2. [x] #57
3. [x] #60
4. [x] #67
5. [ ] Interleaved view support: initial version without optimization (`interleaving-unoptimized`).
6. [ ] Optimization: Retain a cursor to reuse search state when retrieving next row (`cursor-optimization`).
7. [x] Optimization: Reimplement data structures on top of #61 (`stream-port`).

PR #67, opened 2023-03-17
UI testing

martinling:
This PR adds new test infrastructure for recording and replaying UI interactions, along with a couple of test cases.

UI interactions are recorded through normal usage of the UI, by running with the `record-ui-test` feature, which writes out an `actions.json` file describing all actions the user took, and an `output.txt` file containing all information displayed by the UI as a result.

A test case is created by adding a test name to `tests/ui/tests.txt` and creating a directory `tests/ui/<name>`, containing the `actions.json` and the expected output named as `reference.txt`. When the tests are run, the actions are replayed into the UI, and a new `output.txt` file is generated and compared to `reference.txt`.

To the extent I found practical, the replay test exercises all the same UI code as normal operation, including full use of the GTK toolkit, but without actually displaying a window. This raised a couple of side issues:
- On Linux, there must be a valid X `DISPLAY` setting present to use GTK, so CI is modified to run the tests with an Xvfb server available. 
- On MacOS, GTK must run from the main thread. This constraint is incompatible with the standard Cargo test harness, which runs all tests in spawned threads, even if told to limit to one thread.
  - The replay test is therefore implemented as a separate test, `test_replay`, which does not use the standard harness.
  - Since `test_replay` is a separate program, this required moving most of Packetry into a library crate.
  - However, this means that the library target is always built without `cfg(test)` set, so blocks in the UI code that are required by `test_replay` have to be feature gated instead.
  - For this reason, the UI replay test must be included explicitly with `cargo test --features=test-ui-replay`.

Some changes are also made to the existing decoder tests:
  - The list of tests is read from `tests/tests.txt`, rather than listing the `tests` directory.
  - Sequences of identical items are deduplicated, with a single copy and a count included in the output.

This PR is part of the planned sequence documented in #54. The next PR will add interleaving support to the tree view. The  tests in this PR will be used to verify that the UI behaves correctly after the changes.


miek (commented):
This all looks great to me!

One slight issue I'm seeing is that it breaks normal running of `cargo test` (without the test-ui-replay feature) because it doesn't pull in itertools. Adding itertools to `dev-dependencies` without `optional = true` fixed that for me.

martinling:
Oh, nice catch. I think that crept in due to some reordering; later in the #54 sequence itertools becomes a regular dependency so `cargo test` starts working again. Let me figure out which commits to tweak and I'll fix it.

martinling:
Fixed!

miek:
Thanks!

PR #72, opened 2023-04-11
Use new lock-free data structures for capture storage

martinling:
This PR stacks on #61 to reimplement capture storage on top of the new lock-free stream primitive.

First, we add new stream types which build on the basic `stream` implementation:

-  `data_stream` provides storage for arbitrary `Pod` types.
-  `index_stream` provides storage for monotonically increasing values that can be represented as a `u64`.

Next, we make several interdependent changes together in a single commit:

- Replace all `FileVec` instances with data streams.
- Replace all `HybridIndex` instances with index streams.
- Split `Capture` into `CaptureReader` and `CaptureWriter`.
- Split `EndpointTraffic` into `EndpointReader` and `EndpointWriter`.
- Reimplement data structures which must be shared between writers and readers, e.g. `DeviceData`, by using:
   - Atomics from `std::atomic`.
   - The `ArcSwap` and `ArcSwapOption` types.
   - A `SingleWriterRcu` helper trait implementing read-copy-update.
- Remove the `capture` argument from decoder methods, since the decoder can now own the `CaptureWriter` directly.
- Remove the lock calls previously needed to access the capture, since each reader can now have its own read handle.

These changes complete the transition to lock-free capture access, but using uncompressed indexes.

Finally, we add a `compact_index` implementation which reimplements the compression scheme from `HybridIndex` on top of `data_stream` and `index_stream`, and use this to replace the uncompressed indexes.


martinling:
Rebased on main after merge of #61.

PR #75, opened 2023-04-26
Add support for devices to report available analysis speeds

miek:
Based on #73 

This starts off with some refactoring to return a struct with device information from the backend `scan` function, moves querying the device name/description to the backend and returns it in the new struct, adds available speeds to the struct, adds an 'Auto' option, and finally supports the new vendor request to query the available speeds (to be added in https://github.com/greatscottgadgets/luna/pull/202).

### DIFF
--- a/src/backend/luna.rs
+++ b/src/backend/luna.rs
@@ -4,7 +4,13 @@ use std::sync::mpsc::{channel, Sender, Receiver};
 use std::time::Duration;
 
 use num_enum::{FromPrimitive, IntoPrimitive};
-use rusb::{Context, DeviceHandle, UsbContext, Version};
+use rusb::{
+    Context,
+    Device,
+    DeviceHandle,
+    UsbContext,
+    Version
+};
 
 const VID: u16 = 0x1d50;
 const PID: u16 = 0x615b;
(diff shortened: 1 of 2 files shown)


martinling (approved):
Looks good to me. Are you ready to merge #73 too?

miek:
> Looks good to me. Are you ready to merge #73 too?

One thing I noticed with #73 is that, because scan is called during the UI setup, any USB errors stop the UI showing at all (and that becomes more likely with #75 on top). Otherwise it all looks good, so I'm leaning towards merging both and addressing that after if that sounds OK?

martinling:
> One thing I noticed with #73 is that, because scan is called during the UI setup, any USB errors stop the UI showing at all (and that becomes more likely with #75 on top).

That could be fixed fairly easily by removing the `selector.scan()?` line from `DeviceSelector::new`. That would stop USB errors from preventing the UI from coming up. We'd still need to trigger a scan at startup, but we could just schedule that as an idle task at the end of `activate`:

```rust
gtk::glib::idle_add_once(|_| display_error(detect_hardware()));
```

That way any USB error will end up in a dialog box rather than preventing startup.

> Otherwise it all looks good, so I'm leaning towards merging both and addressing that after if that sounds OK?

Fine by me.

PR #77, opened 2023-05-04
More tree model improvements

martinling:
This PR collects some standalone improvements to the `TreeListModel` implementation, in particular:

- Introduce the region map, which simplifies lookups by mapping row number ranges to row data sources (currently either top level items, or the children of some item).
- Introduce the `ModelUpdate` struct, which helps passing around counts of rows to be added, removed and changed.
- Use `u64` rather than `u32` indices in TreeListModel, for better interoperability with the capture database.

These changes are needed for the interleaving work but are independent of it.

### DIFF
--- a/src/tree_list_model.rs
+++ b/src/tree_list_model.rs
@@ -57,9 +57,6 @@ trait Node<Item> {
 
     /// Mark this node as completed.
     fn set_completed(&mut self);
-
-    /// Access this node as an item node, if it is one.
-    fn item_node(&mut self) -> Option<&mut ItemNode<Item>>;
 }
 
 struct Children<Item> {
@@ -172,10 +169,6 @@ impl<Item> Node<Item> for RootNode<Item> {
     fn set_completed(&mut self) {
         self.complete = true;
     }
-
-    fn item_node(&mut self) -> Option<&mut ItemNode<Item>> {
-        None
-    }
 }
 
 impl<Item> Node<Item> for ItemNode<Item> where Item: Copy {
@@ -215,10 +208,6 @@ impl<Item> Node<Item> for ItemNode<Item> where Item: Copy {
                 .remove(&self.item_index);
         }
     }
-
-    fn item_node(&mut self) -> Option<&mut ItemNode<Item>> {
-        Some(self)
-    }
 }
 
 trait UpdateTotal<Item> {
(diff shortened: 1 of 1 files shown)


src/tree_list_model.rs
```diff
@@ -566,21 +978,40 @@ where Item: 'static + Copy,
...
+        if let Ok(position) = u32::try_from(position) {
+            let rows_addressable = u32::MAX - position;
+            let rows_removed = clamp(
+                update.rows_removed + update.rows_changed,
+                rows_addressable);
+            let rows_added = clamp(
+                update.rows_added + update.rows_changed,
+                rows_addressable);
+            model.items_changed(position, rows_removed, rows_added);
+        }
```
antoinevg, line 992: I'm assuming we've got some impedance mismatch with `gtk`'s preference for `u32`'s while our internal model is `u64`?

Is it safe to drop the update here because the position in our model corresponds to the captured data but the position for gtk corresponds to screen position and should never overflow `u32`?

antoinevg (approved):
Code looks good. No crashes or weird behavior observed during execution.

Tested on macOS 13.4 against latest `analyzer` gateware from `cynthion.git` with `--debug`, `--release` and `--features=debug-region-map`.

👍

src/tree_list_model.rs
martinling: > I'm assuming we've got some impedance mismatch with `gtk`'s preference for `u32`'s while our internal model is `u64`?

Yes, that's the basic issue. The idea of commit bc9c8b48a95381e401cd07b72ceed546daa96dac is to put all the handling of that difference into only one place: at the point where we actually implement the `ListModel` interface required by GTK.

> Is it safe to drop the update here because the position in our model corresponds to the captured data but the position for gtk corresponds to screen position and should never overflow `u32`?

It's not a screen position. GTK handles all the business of dealing with what part of the `ListModel` is visible on screen, how to map that to rows of the model, and what rows to retrieve when. The `position` is a row number in the model, not a screen position. But GTK only supports `u32` row indices on `ListModel`.

So the idea is that we achieve compatibility with GTK by presenting it with a model that contains only up to the first 2^32 rows of our internal `u64`-based model.

An update involves a range of rows starting at `position`. If `position` is out of the `u32` range then the whole update is invisible to GTK, and we can skip it. Otherwise, we need to clamp the updated range to include only those rows that are within a `u32`.

This is all theoretical at this point and has not actually been tested, but it is something that could come into play in a capture that's a few hours long.

PR #82, opened 2024-01-05
Use the `anyhow` crate for error handling

martinling:
Use the [anyhow](https://docs.rs/anyhow/latest/anyhow/index.html) crate for error handling.

This crate provides:

- The `anyhow::Error` type, which wraps any other error type, and can thus replace the whole error type hierarchy we have currently including `PacketryError`, `CaptureError`, `StreamError`, `ModelError` etc.

- A `bail!` macro which can simplify sites where we just want to return an error with some text.

- A `Context` trait, implemented for `Option` and `Result`, which allows easily attaching context information when turning an existing `None` or `Err` value into an `anyhow::Error`. This neatly replaces some existing patterns and our `OnBug` trait.

As well as simplifying the code, this change enables us to display a backtrace for any error - this is implemented in the second commit. Backtraces are only captured if `RUST_BACKTRACE` or `RUST_LIB_BACKTRACE` are set.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,24 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the OS, Python version and other tools 
+build:
+  os: ubuntu-22.04
+  tools:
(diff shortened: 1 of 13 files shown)


miek:
I've pushed some CI fixes, could you rebase?

martinling:
Passing CI now.

PR #89, opened 2024-02-08
Port capture backend from `rusb` to `nusb`

martinling:
This PR rewrites the capture backend to use [nusb](https://docs.rs/nusb/latest/nusb/) instead of [rusb](https://docs.rs/rusb/latest/rusb/).

Using `nusb` has the following advantages for us over `rusb`:
- Eliminates a layer of indirection and a dependency on the `libusb` C library: `nusb` uses OS APIs directly.
- Has an API that supports queuing multiple bulk transfers on one endpoint, maximising capture throughput.
- Seems more likely to gain hotplug and RAW_IO support on Windows, both of which are long-stalled in `libusb`.

This PR includes the first usage of `async` in Packetry. I have tried to bring in a minimum of baggage with this:
- [futures-lite](https://docs.rs/futures-lite/latest/futures_lite) for the [`block_on`](https://docs.rs/futures-lite/latest/futures_lite/future/fn.block_on.html) function.
- [futures-channel](https://docs.rs/futures-channel/latest/futures_channel/) for the [`oneshot`](https://docs.rs/futures-channel/latest/futures_channel/oneshot/index.html) channel, to signal the capture task to stop.
- [futures-util](https://docs.rs/futures-util/latest/futures_util/) for the [`select_biased`](https://docs.rs/futures-util/latest/futures_util/macro.select_biased.html) macro and [`FutureExt::fuse`](https://docs.rs/futures-util/latest/futures_util/future/trait.FutureExt.html#method.fuse) method.

The latter two crates are already in our dependency tree as they are used by the `gio` and `glib` crates.

Tested on Linux with Cynthion r1.4.


martinling:
@miek I think I've fixed the issues with this now.

miek (approved):

PR #95, opened 2024-05-30
Proposal for app icon

claui:
I noticed the app doesn’t have an icon yet, so I made you one.


martinling:
Hi @claui, thanks for the proposed icon.

It wasn't something we'd thought about yet, but obviously a GUI app does need an icon, so thanks for taking the initiative!

I shared the icon with the rest of the team at GSG and we had a few thoughts about it:

1. Although the initial use case is USB packet analysis, we'd like to extend the application to other protocols in future, so we'd rather not specifically include a USB connector in the icon.

2. The unique feature of this application, compared to similar tools like Wireshark, is the "packet tree" - hence the name - so it would be nice if the icon could indicate that concept somehow.

3. Some people found the perspective on the packet box to be a bit odd, and suggested an isometric projection would be a better choice.

claui:
Hi @martinling,

Thanks for the thorough feedback!

> Although the initial use case is USB packet analysis, we'd like to extend the application to other protocols in future, so we'd rather not specifically include a USB connector in the icon.

Ok, good to know! Makes sense in that case.  
I’ll remove the connector.

> is the "packet tree" - hence the name

Mind == blown :exploding_head:

Did not see that coming!
Alright, I’ll try out a few ideas and see what works.

> and suggested an isometric projection would be a better choice.

Very good point. I’ll try and figure out the buttons Inkscape wants me to push to make it happen.  
Thanks again for the feedback!

PR #99, opened 2024-06-13
Use mitosis crate to run replay test on macOS

martinling:
WIP on using the [mitosis](https://docs.rs/mitosis/latest/mitosis/) crate to run the UI replay test as a normal unit test on macOS, where GTK must run from the main thread of a process, which `cargo test` does not support.


martinling:
Couldn't get this to work, trying again with `procspawn` on #100.

PR #101, opened 2024-06-14
fix: disable console window for Windows GUI application

nekrondev:
Rust 1.18.0 added support for windows GUI application preventing to open a console window when the application is started.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,3 +1,5 @@
+#![windows_subsystem = "windows"]
+
 use gtk::prelude::*;
 use gtk::gio::ApplicationFlags;
 


nekrondev:
A side note on this: I cross-compiled `Packetry` on Linux using a build container provided by https://github.com/MGlolenstine/gtk4-cross. You need to copy some of the cross compilation environment mingw DLLs to your Windows host but afterwards `Packetry` is running ok (I couldn't yet test with Cynthion so detecting the device using winusb is unknown to me). The build process could be eventually automated to provide an archive including GTK DLLs and `Packetry` executable for easier use on Windows machines.

It looks like this using the standard GTK4 settings provided by the cross compilation environment:

image

martinling:
Thanks for the PR, I hadn't realised this was an issue. I've been testing on Windows, but was always launching the app from the command line, so I didn't notice it would open a console window otherwise.

That cross compilation environment looks useful, but I'd rather avoid that approach in general; we'd like Windows users to be able to contribute to the project easily too, so we want the project to be easy to build on Windows.

I've been doing native builds with MSVC, after getting the dependencies from vcpkg, and just setting a couple of environment variables for the pkg-config stuff.

It took a bit of work to get that working in GitHub Actions (see #83), but the difficulties there were purely on the CI side.

nekrondev:
Ah sorry, I didn't see your PR adding Windows build target, but I agree doing a CI-based native compilation is more useful for developers working on Windows OS. X-compilation also results a bigger executable compared to the native build artifacts found here. Just in case someone wants to create a Windows binary on Linux the docker build container is working fine and took only like 10 minutes to cross compile the whole thing.

PR #112, opened 2024-07-03
Build Windows installer using cargo-wix

martinling:
Builds a Windows MSI installer using [cargo-wix](https://github.com/volks73/cargo-wix), that installs the Packetry application along with all the DLL files required for GTK4 and its dependencies.


martinling:
There's a few things this doesn't do yet, which we might want it to:

- [ ] Sign the installer so that it runs without warnings.
- [ ] Add the application to the Start menu and/or desktop.
- [ ] Associate an icon with the application (see also #95).
- [ ] Associate `.pcap` files with the application.
- [ ] Display the license during installation.
- [ ] Install a copy of the license with the application.

It works though, and could be merged as-is and improved later.

antoinevg (approved):

PR #115, opened 2024-07-06
Update github actions to latest versions

martinling:
- Update `actions/checkout` from v2/3 to v4
- Update `actions/upload-artifact` from v2 to v4
  - Avoid conflicting artifacts by uploading only those from the `stable` Rust build.
- Replace unmaintained `actions-rs` actions with direct use of `rustup` and `cargo`.
- Always use the latest version of the Ubuntu runner.
- Replace deprecated `GabrielBB/xvfb-action` with direct use of `xvfb-run`.
- Don't try to install `pkg-config` on macOS; the runner has it.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -16,7 +16,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - uses: awalsh128/cache-apt-pkgs-action@latest
         with:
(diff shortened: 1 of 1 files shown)


martinling:
I was hoping this would fix the slew of deprecation warnings that show up in each of our CI runs, but it turns out most of those are due to unmaintained actions from the now-archived @actions-rs organisation.

So we need to see about replacing those too.

martinling:
Yeah, I think `actions-rs/toolchain` and `actions-rs/cargo` can be replaced easily by direct use of `rustup` and `cargo.

I'm not sure what to do about `actions-rs/clippy-check` - there are several forks/successors.

miek:
I was thinking just run `cargo clippy` directly too. It seems like the fancy annotations that the action does often fail on PRs anyway & having the normal output from clippy would be more useful.

martinling:
All done. No more warnings!

PR #124, opened 2024-07-12
Add initial user documentation

mossmann:
Readthedocs preview: https://packetry--124.org.readthedocs.build/en/124/


docs/source/quick_start.rst
```diff
@@ -0,0 +1,140 @@
...
+
+   Run the Packetry executable.
+
+.. tab:: Cargo
+
+   In the ``packetry`` directory run:
+
+    .. code::
+
+        cargo run
```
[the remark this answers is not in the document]
martinling: It's probably worth mentioning here that they can use `cargo install --path .` to build and install the binary to their `.cargo/bin` directory, which will mean running `packetry` will work from anywhere if that's in their `PATH`. 

Or, after running `cargo build --release`, they can copy `target/release/packetry[.exe]` to somewhere else in their `PATH`, like `/usr/local/bin` on Unix-like systems.

docs/source/quick_start.rst
```diff
@@ -0,0 +1,140 @@
...
+        .. code::
+
+            cd packetry
+            cargo build
+
+        Note: Do not build with ``--all-features``. All the optional features currently in the package are for debug/test purposes only, and will prevent or degrade normal use of the application if enabled. See ``Cargo.toml`` for details.
+
+.. tab:: Windows
+
+   Download and run the Windows installer from the `latest release <https://github.com/greatscottgadgets/packetry/releases/latest>`__.
```
martinling, line 98: We should also give instructions for building from source on Windows. We want Windows-based developers to be able to contribute easily.

A suitable workflow would be something like:
```
# Get vcpkg and set it up.
git clone https://github.com/microsoft/vcpkg.git
cd vcpkg
bootstrap-vcpkg.bat

# This will take a while - it builds and installs GTK and all its dependencies from source.
vcpkg install gtk pkgconf

# Set environment variables so that Packetry can find its dependencies. These will only
# persist for the lifetime of the current command window, so you may want to set these
# up elsewhere for future use.
set PKG_CONFIG=%CD%\installed\x64-windows\tools\pkgconf\pkgconf.exe
set PKG_CONFIG_PATH=%CD%\installed\x64-windows\lib\pkgconfig
set PATH=%PATH%;%CD%\installed\x64-windows\bin
cd ..

# Get Packetry and build it.
git clone https://github.com/greatscottgadgets/packetry.git
cd packetry
cargo build --release
```

docs/source/user_interface.rst
```diff
@@ -0,0 +1,54 @@
...
+- **Open Button**: Loads a ``.pcap`` file with the ``LINKTYPE_USB_2_0`` link-layer header type.
+- **Save Button**: Saves a ``.pcap`` file with the ``LINKTYPE_USB_2_0`` link-layer header type.
+- **Scan Button**: Searches for a capture device such as a Cynthion running analyzer gateware.
+- **Capture Button**: Starts capturing USB data with the connected capture device.
+- **Stop Button**: Stops capturing USB data.
+- **Device Drop-down**: Selects a capture device. If you don't see a device listed here, try using the Scan button.
+- **Speed Drop-down**: Selects the target device speed.
+
+
+Traffic Window
```
martinling, line 22: I would call this the "Traffic View" - it's not a separate window.

docs/source/user_interface.rst
```diff
@@ -0,0 +1,54 @@
...
+
+.. image:: ../images/traffic-window.png
+  :alt: Traffic Window
+
+The Traffic Window displays packets in an interactive hierarchical view, allowing you to expand high level events to explore their constituent low level transactions and packets.
+
+Each packet is displayed with a precise time stamp relative to the start of capture. Higher level groups have a time stamp matching the first packet that is a part of the group.
+
+
+Device Window
```
martinling, line 35: Similarly, the "Device View" - it's not a separate window.

martinling (changes requested):

docs/source/conf.py
```diff
@@ -17,14 +17,15 @@
 # -- General configuration ---------------------------------------------------
```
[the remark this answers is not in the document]
mossmann: Wouldn't that update the date any time someone builds the documentation? If our last update was in 2028 and someone builds it in 2035, it would be incorrect to list a copyright of 2035.

docs/source/user_interface.rst
mossmann: I had wondered about this. I took the name "Window" from the source code, but I was thinking something like "View", "Panel", "Pane", or "Area" might be better. Might "View" conflict with whatever we call it when we provide one or more additional ways to display traffic (interleaved tree or strict chronological packets)?

docs/source/conf.py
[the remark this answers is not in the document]
antoinevg: Well _technically_ the copyright text for the release would then be unique and _could_ be considered part of the copyrighted material… 😂 but that's a good point. 🙈

docs/source/user_interface.rst
martinling: It has that name in the source code only because it's a `Window` internally to GTK. It's not a window in terms of how it's presented to the user as a UI element.

I agree that "View" isn't the best choice, because we're going to offer different views within that same space.

I think "Pane" is the right choice. That matches what GTK calls those draggably-divided areas created by [Gtk.Paned](https://docs.gtk.org/gtk4/class.Paned.html).

mossmann: Okay, I'll go with "Pane". I had a look through documentation of some popular open source GUI applications (Inkscape, Gimp, LibreOffice) and saw "Pane" used once or twice.

docs/source/quick_start.rst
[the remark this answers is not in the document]
mossmann: `cargo install --path .` seems to me like it *should* be the best option for folks who have used `rustup`, but it fails in an unfriendly way for folks who haven't. #sigh#

How about this (with the hint)?
```
cargo build --release
sudo cp target/releases/packetry /usr/local/bin
```
Hint: You can copy the `packetry` executable to any directory in your `PATH` instead of `/usr/local/bin`. If you've used `rustup` and have `.cargo/bin` in your `PATH`, `.cargo/bin` is a great choice!

docs/source/quick_start.rst
```diff
@@ -103,7 +103,56 @@ Install Packetry
...
+
+        **Build and Install Packetry**
+
+        .. code:: bat
+
+            cd packetry
+            cargo build --release
+            set PATH=%PATH%;%CD%\target\releases
+
+        This ``PATH`` modification will only persist for the lifetime of the current command window, so you may want to set it up elsewhere for future use. Alternatively you can copy ``target\releases\packetry`` to a location already in your ``PATH``.
```
martinling, line 153: It's `target\release`, not `target\releases`.

martinling (changes requested):

docs/source/quick_start.rst
mossmann: oops! fixed.

docs/source/quick_start.rst
```diff
@@ -0,0 +1,191 @@
...
+Install Packetry
+----------------
+
+.. tab:: Linux
+
+    .. tab:: Binary
+
+        AppImage coming soon! For now you can build and install with Cargo.
+
+    .. tab:: Cargo
```
martinling, line 26: I feel like "Pre-built" and "From source code" would be friendlier headings than "Binary" and "Cargo" - both here and in the other OS sections.

docs/source/quick_start.rst
```diff
@@ -0,0 +1,191 @@
...
+.. tab:: Windows
+
+    .. tab:: Binary
+
+        Download and run the Windows installer from the `latest release <https://github.com/greatscottgadgets/packetry/releases/latest>`__.
+
+    .. tab:: Cargo
+
+        **Install Prerequisites**
+
```
martinling, line 113: I suggest we add a note to indicate that there are other ways one can get the prerequisites as we don't specifically require vcpkg:
```suggestion

Building Packetry requires that the GTK 4 libraries and header files are available, and discoverable via the `pkg-config` mechanism. There are several ways you can arrange this; if in doubt we suggest using [vcpkg](https://vcpkg.io/en/) as follows:

```

martinling (approved):
Looks good! I suggested a couple of minor changes but they're not essential.

docs/source/quick_start.rst
mossmann: Is there a reason to treat this differently than the other "Install Prerequisites" sections? We're documenting our recommended and supported process, not every possible way to do things.

mossmann: I was thinking that we may later add a Development section to the documentation and potentially move all the "From source code" installation instructions and most of the current README contents there. Then we could make those instructions more verbose, and the quick start installation instructions could direct folks to the Development section if they prefer to install from source.

martinling: I think there is a good reason to mention this specifically on Windows, which is that there are a lot of projects that are very tied into the vcpkg ecosystem (e.g. have vcpkg as a submodule, and require it to build successfully). I'd like it to be clear that we're not one of those projects; as long as people have GTK discoverable by `pkg-config` somehow, things should work. That means that e.g. people who already have MSYS2 set up can just use that instead.

martinling: To be honest, I'm not really sure if vcpkg is the best thing to recommend in the first place. It requires building everything from source, which takes a lot of time and disk space. I'm not keen on recommending MSYS2 though, because it's a lot more more invasive whereas vcpkg is self-contained.

Another option would be to suggest the binary packages from https://github.com/wingtk/gvsbuild, but I haven't tested that option myself yet.

antoinevg (approved):
👍

PR #125, opened 2024-07-12
Add `--version` option and About dialog

martinling:
This PR adds a `--version` command line option, and an About dialog to the UI.

Version information is captured at compile time with the [git_version](https://docs.rs/git-version/latest/git_version/) macro, so that it is possible to identify builds from non-release versions.

The About dialog includes basic package information, the BSD license text for Packetry itself, and some system information: OS, architecture and GTK library version in use.

It does not yet include copyright or license information for any dependencies.

PR #126, opened 2024-07-13
Gather more build information and improve version string

martinling:
This PR uses the [`built`](https://docs.rs/built/latest/built/) crate to gather additional build-time information which is reported in the `--version` output and the About dialog.

Fixes the failure to build without a `.git` directory, caused by #125.

Output of `--version` when built with a `.git` directory:
```
Packetry version 0.1.0 (git b2029df)

Runtime information:
  GTK version: 4.12.5

Packetry build information:
  Git commit: b2029df15600f1ddcb2e0655b6baa0abfc6355cd
  Cargo package version: 0.1.0
  Enabled features: (none)

Rust compiler:
  Version: rustc 1.79.0 (129f3b996 2024-06-10)
  Target: x86_64-unknown-linux-gnu (little-endian, 64-bit)
  Optimization level: 3
  Debug build: no
```

Output when built without a `.git` directory:
```
Packetry version 0.1.0

Runtime information:
  GTK version: 4.12.5

Packetry build information:
  Git commit: unknown
  Cargo package version: 0.1.0
  Enabled features: (none)

Rust compiler:
  Version: rustc 1.79.0 (129f3b996 2024-06-10)
  Target: x86_64-unknown-linux-gnu (little-endian, 64-bit)
  Optimization level: 3
  Debug build: no
```

The `--dependencies` option can now be used along with `--version` to print crate dependencies used. This is omitted by default as it generates a lot of output.

The same information as above, as well as the dependency listing, is displayed in the About dialog.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -94,6 +94,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf4b9d6a944f767f8e5e0db018570623c85f3d925ac718db4e06d0187adb21c1"
 
 [[package]]
+name = "built"
+version = "0.7.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "236e6289eda5a812bc6b53c3b024039382a2895fbbeef2d748b2931546d392c4"
+dependencies = [
+ "cargo-lock",
+ "git2",
+]
+
+[[package]]
 name = "bytemuck"
 version = "1.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


mossmann:
I tried it with local tags and observed that this results in `git describe`-style version strings like:

```
Packetry version 0.1.0 (git v0.0.1-8-g96882e5)
```

or
```
Packetry version 0.1.0 (git v0.1.0)
```

which I think is great!

mossmann:
Is it expected that I would see things like `windows_aarch64_msvc 0.48.5` in my dependency list on Linux/x86_64?

martinling:
> I tried it with local tags and observed that this results in `git describe`-style version strings

Yes, that was the intent and I tested it too; I should have mentioned that.

> Is it expected that I would see things like windows_aarch64_msvc 0.48.5 in my dependency list on Linux/x86_64?

It's expected, but annoying. It happens because we depend on cross-platform crates that depend on the union of all their dependencies for each platform.

Originally, they had no option but to do this. Later, [Cargo gained support for target-specific dependencies](https://github.com/rust-lang/cargo/issues/7914). But some projects haven't been updated to use that feature.

mossmann (approved):
Love it! The additional info is excellent!

martinling:
Actually, looking a bit more closely that may not be the reason for the extraneous dependencies. See https://github.com/lukaslueg/built/issues/55. So we may be able to do better here.

martinling:
Updated to catch the license file for a dependency after fixing #127; no change to the diff of the main commit.

PR #148, opened 2024-07-30
Fixed capture save not working

Matt-Allen-Bose:
Currently when you try to save a capture an error is thrown from GIO File info query since the pcap doesn't exist yet. I've moved the file info into the Load action to fix this and removed the ui.filename assignment as it doesn't seem to be needed for saving/reading.


miek:
Good catch, thanks!

PR #154, opened 2024-07-31
Add `packetry-cli` wrapper program

martinling:
This is necessary on Windows, where we can't have the same binary serve as both a GUI and a console program, due to the constraints of the PE binary format. See e.g. [this post by Raymond Chen](https://devblogs.microsoft.com/oldnewthing/20090101-00/?p=19643), and [this StackOverflow discussion](https://stackoverflow.com/questions/493536/can-one-executable-be-both-a-console-and-gui-application).

Instead, we add a new `packetry-cli` console binary which simply acts as a wrapper around the main GUI binary.

The wrapper sets the `PACKETRY_ATTACH_CONSOLE` environment variable, which tells the main process to attach to the wrapping parent's console. The wrapper waits for the child to complete, and mirrors its exit code.

Thanks to @pointy56 for investigating and prototyping this solution.


martinling:
Note: Cargo doesn't currently support platform-specific binary targets. So we have to build `packetry-cli` on the other operating systems too, even though it's a no-op wrapper.

I think this is the right choice though, because it means that scripts and documentation can be written to always use `packetry-cli` for command line operations, and everything should then work unmodified on all systems.

mossmann (changes requested):
I think this is a great solution! I tested on Linux, macOS, and Windows, and everything worked except that I found a bug with `--test-cynthion --save-captures`. On Windows I got:
```
Capture enabled, speed: High (480Mbps)
Starting read from test device
Read 4096 bytes from test device
Requesting capture stop
Capture disabled
thread 'main' panicked at src\test_cynthion.rs:28:64:
called `Result::unwrap()` on an `Err` value: The system cannot find the file specified. (os error 2)
```
On macOS I saw similar ("No such file or directory"). I retested on macOS with Packetry 0.1.0 from Homebrew, and `--save-captures` worked. I haven't tested `--save-captures` on Linux.

mossmann (approved):
I reproduced the bug on Linux and with the main branch, so it was not introduced with this PR. I'll open an issue.

PR #161, opened 2024-08-06
Add detail pane

### LINKED ISSUE #10: Add a detail pane to GTK frontend
This would show more detail about the currently selected `Item`, similar to the behaviour in ViewSB or Wireshark

### LINKED ISSUE #156: Add word wrap for longer data with optional ASCII decoding
It would be nice to have the Packetry to wrap the data in the expanded view and also add option for automatic ASCII decoding. When longer data packets are captured, you cannot inspect them (unless you have really wide monitor).

image

martinling:
Adds an initial implementation of the detail pane, populated with a more detailed, multi-line description of whichever item is selected in the traffic view.

Unlike the summaries in the tree view, the detail pane will display up to 1024 bytes of payload data, wrapping across lines as required, and it is possible to select and copy from the text shown.

This is not a complete solution for the detail pane yet, as it doesn't address viewing the data for transfers of >1024 bytes. For this we'll need a solution based on `ListView` that loads data on demand, as the transfer size is unbounded.

However, it gets the new pane into the UI, and addresses the immediate need for a way to access the full content of packets and transactions.

Closes #10
Closes #156


martinling:
Revised slightly to avoid using `fmt_count`, which adds thousand separators, for packet and transaction IDs.

I'm still using it for timestamps, because it helps separate milli/micro/nanoseconds visually, but I don't think it's helpful for IDs. Keeping those as just digits makes it easier to e.g. paste them into Wireshark's 'Go to packet' input.

miek (approved):

PR #162, opened 2024-08-08
test: remove unused import in test cfg

jk-ozlabs:
When running tests we have an unused import:

    warning: unused import: `self`
      --> src/ui.rs:24:17
       |
    24 | use gtk::glib::{self, Object, SignalHandlerId};
       |                 ^^^^
       |
       = note: `#[warn(unused_imports)]` on by default

The one use of the `self` import is not(cfg(test)); all other uses reference the full gtk::glib path.

Change the glib (::self) use to suit the non-test uses.

### DIFF
--- a/src/ui.rs
+++ b/src/ui.rs
@@ -21,7 +21,7 @@ use gtk::gio::{
     MenuItem,
     SimpleActionGroup
 };
-use gtk::glib::{self, Object, SignalHandlerId};
+use gtk::glib::{Object, SignalHandlerId};
 use gtk::{
     prelude::*,
     AboutDialog,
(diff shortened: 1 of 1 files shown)


src/ui.rs
```diff
@@ -1055,7 +1055,7 @@ fn show_about() -> Result<(), Error> {
 pub fn display_error(result: Result<(), Error>) {
     #[cfg(not(test))]
     if let Err(e) = result {
-        if let Some(g_error) = e.downcast_ref::<glib::Error>() {
+        if let Some(g_error) = e.downcast_ref::<gkt::glib::Error>() {
```
martinling, line 1058: Typo here.

martinling:
Merging with the typo fixed.

src/ui.rs
jk-ozlabs: bah, I was sure I had built that, evidently not. Thanks for the fix.

PR #163, opened 2024-08-08
ui: Add bulk/int/iso/cntl type to endpoint summary info

jk-ozlabs:
The Endpoint summary line in the Interface descriptor list currently contains the address and direction data:

  Endpoint 3 IN

This change adds the endpoint type too, as derived from the attribute data:

  Endpoint 3 BULK IN

\- by defining std::fmt::Display for EndpointType, and using that in the summary data.

### DIFF
--- a/src/capture.rs
+++ b/src/capture.rs
@@ -1637,12 +1637,12 @@ impl ItemSource<DeviceItem> for CaptureReader {
                 interface.descriptor.field_text(*field, strings.as_ref())
             },
             EndpointDescriptor(dev, conf, iface, ep) => {
-                let addr = self.device_data(dev)?
-                               .configuration(conf)?
-                               .interface(iface)?
-                               .endpoint_descriptor(ep)?
-                               .endpoint_address;
-                format!("Endpoint {} {}", addr.number(), addr.direction())
+                let config = self.device_data(dev)?.configuration(conf)?;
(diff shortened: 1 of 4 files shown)


jk-ozlabs:
Bikeshed opportunity: should it be `CNTL` or `CTRL`?

jk-ozlabs:
With this, the endpoint info looks like:

bulky

martinling:
Nice! This is a good addition, thank you.

I don't think it's necessary to capitalise or abbreviate the endpoint types, though. Just `bulk`, `interrupt`, `isochronous` or `control` should be fine here, and that matches how we label transfers in the traffic view.

We capitalise `IN` and `OUT` because those direction identifiers are always specifically capitalised in the USB 2.0 standard.

It would also make more sense to put the number and direction together. The direction and the endpoint number together form a single value, the endpoint address. There can be both a `1 IN` (address 0x81) and `1 OUT` (address 0x01).

Overall I think what I'd prefer to see for your example is:

Endpoint 1 IN (interrupt)
Endpoint 2 OUT (bulk)
Endpoint 3 IN (bulk)

jk-ozlabs:
Sure thing, can do. I'm not sure where I saw the `BULK IN` endpoint naming formats previously (or whether I'm just imagining that as a convention!)

jk-ozlabs:
With ffa5061b1:

bulky

martinling:
Great, thank you!

PR #166, opened 2024-08-08
Select dark theme if detected as a system preference.

### LINKED ISSUE #121: Packetry does not follow system-level dark mode on Windows and macOS
On Windows and macOS, Packetry appears in a light theme even if dark mode is selected globally in the OS.

It's not clear to me yet how this is supposed to work - I'd have thought it was GTK's responsibilty and shouldn't require code in the application.

Perhaps we'd need to use [`libadwaita::Application`](https://world.pages.gitlab.gnome.org/Rust/libadwaita-rs/stable/latest/docs/libadwaita/struct.Application.html) instead of `gtk::Application`? But that seems wrong, as we're trying to produce a generic cross-platform application, not a GNOME one.

martinling:
Uses:
- Rust [dark-light](https://github.com/frewsxcv/rust-dark-light) crate to detect the system setting
- GtkSettings [set_gtk_application_prefer_dark_theme](https://gtk-rs.org/gtk4-rs/stable/latest/docs/gtk4/struct.Settings.html#method.set_gtk_application_prefer_dark_theme) to configure the GTK behaviour.  

Closes #121.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -39,7 +39,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ['macos-latest', 'ubuntu-latest', 'windows-latest']
-        rust: ['stable', '1.74']
+        rust: ['stable', '1.75']
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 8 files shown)


martinling:
Tested successfully on Windows

image

and macOS

Eak00hB

martinling:
There were a few issues adding the `dark-light` crate as a dependency which are worth noting here:

- This has the side effect of bumping our minimum supported Rust version from 1.74 to 1.75, not due to `dark-light` itself but to one of its dependencies. I think this makes very little difference for us in practice.

- The `dark-light` crate is missing a license file and copyright notice, but their license is clearly specified in the README so I have added it manually with copyright lines for the two primary maintainers.

- The `malloc_buf` crate is missing the LICENSE file in the crate download, but it's present in the repository, so I've added it manually.

- One of the new dependencies upset the Windows installer build, because of an author name with characters that are present in UTF-8 but not in Windows code page 1252, which is the default encoding for stdout there. Fixed by passing the `-X utf8` option to Python.

PR #181, opened 2024-08-27
Keyboard shortcuts

### LINKED ISSUE #135: Add keyboard shortcuts
We should add keyboard shortcuts for common operations. Some obvious ones would be:

- Open file (Ctrl-O, or ⌘-O on Mac)
- Save file (Ctrl-S, or ⌘-S on Mac)
- Start capture
- Stop capture

Specific suggestions would be welcome.

miek:
This adds keyboard shortcuts for common actions. Fixes #135

Initially I tried connecting them up to call Button::emit_clicked, which would've been a bit simpler, but I had issues with borrowing `UI` to get at the button & the callbacks needing to borrow it too (it also felt like a bit of a hack). So instead I've moved all the callbacks over to [actions](https://gtk-rs.org/gtk4-rs/stable/latest/book/actions.html) which seems to be the preferred way. This also required moving the logic for stop behaviour into a function rather than connecting/disconnecting signals.

For the shortcuts themselves, I've used conventional ones where possible. For scan I used Ctrl+R and F5, to mirror common browser refresh shortcuts. For capturing, I went for Ctrl+b and Ctrl+e, for begin and end respectively. I'm open to suggestions for anything different though.

I'm leaving this as draft for now as I've just remembered I need to handle the macOS keys properly, but otherwise it's open for any feedback.


src/ui.rs
```diff
@@ -356,27 +378,55 @@ pub fn activate(application: &Application) -> Result<(), Error> {
...
         .build();
     let capture_button = gtk::Button::builder()
         .icon_name("media-record")
         .tooltip_text("Capture")
+        .action_name("win.capture")
         .build();
     let stop_button = gtk::Button::builder()
         .icon_name("media-playback-stop")
         .tooltip_text("Stop")
+        .action_name("stop")
```
martinling, line 429: Shouldn't this be `win.stop`?

miek: Yep, thanks!

miek:
> I found that it was possible to simplify this quite a bit, see [this commit](https://github.com/martinling/packetry/commit/31a17590ec77d69bf17dc4a0e3a8c9663e6e4d02).
Thanks! I like it. I've gone ahead and rebased everything with that change merged into the first commit.

src/ui.rs
```diff
@@ -1015,12 +1058,22 @@ fn save_pcap(file: gio::File,
...
-        ui.scan_button.set_sensitive(true);
+        match std::mem::replace(&mut ui.stop_state, StopState::Disabled) {
+            StopState::Disabled => {},
+            StopState::Pcap(cancel_handle) => {
+                STOP.store(true, Ordering::Relaxed);
+                cancel_handle.cancel();
+            },
+            StopState::Cynthion(stop_handle) => {
+                stop_handle.stop()?;
+                ui.scan_button.set_sensitive(true);
```
martinling, line 1071: This line can be removed, it's duplicated below (my bad!).

martinling:
This needs some documentation so that people can find out about the keyboard shortcuts.

Better still would be some indication of them within the UI - maybe a new item in the dropdown menu that leads to the About dialog?

miek:
> In fact GTK has a built-in [`ShortcutsWindow`](https://docs.gtk.org/gtk4/class.ShortcutsWindow.html) for this.

I did have a brief look into that, but found https://github.com/gtk-rs/gtk4-rs/discussions/1724 which points out that it'd need a very recent version of GTK to work without using XML to drive it.

martinling:
Ah, right. Doing it with XML [doesn't look too awful](https://github.com/gtk-rs/gtk4-rs/blob/master/examples/gtk_builder/main.rs), the XML file can be embedded into the binary at build time.

I'm fine with not pursuing that just now though as long as the keyboard shortcuts are added to the docs instead.

miek:
I've added the keyboard shortcuts to the docs ([rendered here](https://packetry--181.org.readthedocs.build/en/181/)). I figured it was worth listing the linux/windows and mac ones separately for clarity and in case they diverge later. I tried using the inline-tab interface (like we use for build/install instructions) but it doesn't let you have sub-headings within it, so I just made them separate pages.

martinling (approved):
That looks fine!

PR #185, opened 2024-09-10
Initial PcapNG file format support

martinling:
This draft PR adds initial support for the [PcapNG](https://pcapng.com/) file format.

The following features are supported:
- Saving captures to a file with:
    - Native endianness
    - A single section header
    - A single interface with the `USB_2_0` link type only
    - Nanosecond timestamp resolution
- Loading captures from files with:
    - Any endianness
    - A single section header
    - A single interface with one of the following link types:
        - `USB_2_0`
        - `USB_2_0_HIGH_SPEED`
        - `USB_2_0_FULL_SPEED`
        - `USB_2_0_LOW_SPEED`
    - Any timestamp resolution


martinling:
Added support for metadata:

- Storing metadata about the current capture in a `CaptureMetadata` type.
- Loading and saving metadata in PcapNG files.
- Generating appropriate metadata for Cynthion captures.
- A UI dialog for viewing the capture metadata and editing the comment field.

Some further steps here are blocked on external issues:

1. It's not currently possible to save the comment field in the section header, due to a lifetime issue in the `pcap-file` crate, which I have submitted a fix for:
https://github.com/courvoif/pcap-file/pull/41

2. I'd like the user to be able to select between the generic `USB_2_0` link type and the speed-specific e.g. `USB_2_0_HIGH_SPEED` link types when saving, as applicable to the current capture. This choice applies to both the Pcap and PcapNG formats, so would be best offered as an extra dropdown in the save dialog. This is blocked on a gtk4-rs bug which I've had a fix merged for, but we're still waiting on a new release before we can use it:
https://github.com/gtk-rs/gtk4-rs/pull/1834

martinling:
My gtk4-rs fix is now released in [0.9.2](https://github.com/gtk-rs/gtk4-rs/releases/tag/0.9.2).

martinling:
We now have [our own fork of the `pcap-file` crate](https://crates.io/crates/pcap-file-gsg), and this branch has been updated to depend on it.

This allows removal of a lot of code which is now implemented in the library instead.

PR #188, opened 2024-09-12
Handle descriptors that are longer than defined in the specification.

x0rloser:
Handle descriptors that are longer than defined in the specification.
Follow specification guidelines for doing so.

While testing packetry I came across descriptors that were longer than shown in the usb specification. Currently packetry skips over them when they are not the exact same size as in the specification. However according to the specification they are valid and the extra data bytes should just be discarded. This fix implements this.

I tested this fix with a packet log I had previously captured and it now handles the descriptors correctly.

Reference:
Universal Serial Bus Specification (Revision 2.0)
usb_20.pdf - page 260

"If a descriptor returns with a value in its length field that is less than defined by this specification, the descriptor is invalid and should be rejected by the host. If the descriptor returns with a value in its length field that is greater than defined by this specification, the extra bytes are ignored by the host, but the next descriptor is located using the length returned rather than the length expected."

(I just started learning rust today - so I'm open to any required changes)


martinling:
This change looks correct, but I would like to add a test case for this behaviour.

Do you have a capture that you can share which includes a longer-than-expected descriptor?

PR #196, opened 2024-10-09
Add context menu system and options to save data

martinling:
Allows the user to right click on a transfer, select "Save payload data...", choose a filename and save the payload.


martinling:
A whole farm of yaks have now been shaved, and as a result this is now complete and ready for review.

It supports saving payload data from data transfers, control transfers, transactions and packets from the traffic view.

Descriptor data can also be saved from the device view.

miek (approved):

PR #197, opened 2024-10-10
Display all descriptor types in device view

martinling:
This PR extends the device view to display additional descriptor types.

- Interface Association Descriptors are now decoded, and their interfaces grouped into a Function.
- All descriptors are now shown, even where not understood by Packetry.
- Arbitrary descriptors are shown attached to configurations, interfaces or endpoints, according to where they appear in the configuration data.
- Functions and interfaces are displayed with their associated class, based on `usb.ids` data.

Example:
image

A subsequent PR will support saving descriptors to files for further analysis, building on the features in #196.


miek:
Could you rebase this? There's just a minor conflict from merging #193

martinling:
Done!

martinling:
I've just realised that one thing in the last commit might need some further thought.

I shortened the notation `Interface 1 (alternate 2)` to `Interface 1.2`, to make more space for the interface class beside it.

However, in the traffic view we use that same notation to indicate <device number>.<interface>, e.g. in this request description:

`Setting alternate setting 0 for interface 25.2`

So perhaps we should go with something else like `Interface 1 alt 2: Communications` in the device view.

martinling:
I've gone ahead and made the change to `Interface 1 alt 2` etc in the last commit.

PR #199, opened 2024-10-11
Fix open button not being re-enabled after stop

### LINKED ISSUE #198: Open button doesn't get re-enabled after opening a file or running a capture
I think the call to re-enable it got missed in 1f5c359

martinling:
Fixes #198.


miek:
Thanks!

PR #218, opened 2025-02-12
actions: add runner version to apt cache key

miek:
This adds the actions runner image version to the apt cache key, so it should get rebuilt when the runner is updated.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -15,13 +15,26 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Get runner version
+        run: |
+          # There is currently no way to access the runner version nicely,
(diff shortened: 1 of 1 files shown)


martinling (approved):
Bodgetastic.

Note that if we ever change the set of packages to be installed, we'll need to add our own value to the cache key as well, so that we can bump it without the runner version having changed.

But we can just do that when we first need it.

miek:
Ahh, I was wondering why the version was already set - that makes sense. Though it looks like it already adds the list of packages to the hash anyway (https://github.com/awalsh128/cache-apt-pkgs-action/blob/master/pre_cache_action.sh#L61-L73) so it might be fine?

martinling:
Oh cool, if it's already including that in the hash then we shouldn't need to do anything. :+1:

PR #228, opened 2025-03-27
Add support for dumping & restoring capture database

martinling:
This PR adds code to support dumping the capture database to disk, and restoring it back.

This is intended for debugging purposes only. The dump format is hierarchical, with data structures mapped to directories and files.

To verify that this works correctly, the `test_captures` test is modified to dump and restore the capture database in between decoding each capture and querying it for expected results.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /docs/build/
 /target
+/tests/*/dump
 /tests/*/output.txt
 /tests/*/devices-output.txt
 /tests/ui/*/output.txt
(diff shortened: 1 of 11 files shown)


antoinevg (approved):
Looking good. 

`cargo test`ed on Linux and macOS. 

(Had a bit of trouble during compilation with `libfuzzer` on macOS but that was sorted out by re-installing the Xcode command-line tools.)

martinling:
Thanks for the review!

PR #230, opened 2025-04-03
Additional clippy fixes.

martinling:
More nitpicks for our paperclip-shaped friend.

In the process, I found a [bug in clippy](https://github.com/rust-lang/rust-clippy/issues/14531) !


martinling:
Now with even more paperclippy goodness, because 1.86.0 dropped.

PR #234, opened 2025-05-22
WIP: Database snapshotting

### LINKED ISSUE #213: Off-by-one bug: Requested read of range one larger than stream length
Perhaps there's an off by one bug somewhere, I can't save my capture to a file at all with v0.4.0 (appimage).
error

martinling:
Work in progress.

Rather than trying to work with live data from the database during UI updates, take a snapshot of the database state and then work with that snapshot until the next update. This eliminates a lot of potential for non-determinism that's otherwise hard to test or reproduce.

The challenge is to make the snapshotting cheap; this can be achieved by grouping all the `AtomicU64` counters used by the database into a contiguous block of memory which serves as the state representation, and which can be snapshotted effectively with a `memcpy` by the writing thread, in between handling packets.

Further work is still needed to make constructing and working with snapshots more efficient.

Fixes some crashes in #150.

Probably fixes #213 and similar occasional failures.

### DIFF
--- a/src/capture.rs
+++ b/src/capture.rs
@@ -78,6 +78,7 @@ pub struct CaptureShared {
 pub struct CaptureWriter {
     pub counters: CounterSet,
     pub shared: Arc<CaptureShared>,
+    pub endpoint_writers: VecMap<EndpointId, EndpointWriter>,
     pub packet_data: DataWriter<u8, PACKET_DATA_BLOCK_SIZE>,
     pub packet_index: CompactWriter<PacketId, PacketByteId, 2>,
     pub packet_times: CompactWriter<PacketId, Timestamp, 3>,
@@ -144,6 +145,7 @@ pub fn create_capture()
     let writer = CaptureWriter {
         counters,
         shared: shared.clone(),
+        endpoint_writers: VecMap::new(),
         packet_data: data_writer,
         packet_index: packets_writer,
         packet_times: timestamp_writer,
--- a/src/capture.rs
+++ b/src/capture.rs
@@ -5,13 +5,14 @@ use std::fmt::{Debug, Write};
 use std::iter::once;
 use std::num::NonZeroU32;
 use std::ops::Range;
-use std::sync::atomic::{AtomicBool, AtomicU32, AtomicU64};
+use std::sync::atomic::{AtomicBool, AtomicU32};
 use std::sync::atomic::Ordering::{Acquire, Release};
 use std::sync::Arc;
 use std::time::Duration;
 use std::mem::size_of;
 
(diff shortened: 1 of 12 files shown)


miek:
I'm getting a new error now :slightly_smiling_face: 
If I connect up a HackRF for live capture, then hit reset on it during the capture, I get messages like "requested position 49642 but snapshot length is 49153".

martinling:
I've made some progress on debugging this, and I have a workaround, though not yet a proper fix.

The crash was happening when `TreeListModel` tried to update the text description of an item that had gained children since the last update.

When we first call `ItemSource::description` for an item, we catch errors and display them in the list view, where the text would have been. This avoids a crash, and is also really helpful for debugging the code that produces the descriptions, because it's possible to examine the problematic item in-situ and look at its contents.

However, when we _update_ a description because something about the item has changed, we weren't doing that. So the workaround is just:

```diff
             if item_updated {
                 // The node's description may change.
-                let summary = cap.description(&item_node.item, false)?;
+                let summary = match cap.description(&item_node.item, false) {
+                    Ok(description) => description,
+                    Err(e) => format!("Error: {e}"),
+                }
```
With that error handling, everything works fine from the user's perspective, and I've not seen any errors actually become visible in the UI. However, adding some further code to log the errors, along with the _previous_ description of the items in question, reveals what's failing:
```
Failed updating '13840 SOF packets': requested position 16123 but snapshot length is 15582
Failed updating '20 SOF packets': requested position 17469 but snapshot length is 17389
Failed updating '12081 SOF packets': requested position 30938 but snapshot length is 30124
Failed updating '9874 SOF packets': requested position 43503 but snapshot length is 42777
Failed updating '8762 SOF packets': requested position 56080 but snapshot length is 55767
```
The failed lookup is into the `packet_data` stream, and comes from a `packet_pid()` call in `CaptureReaderOps::transaction(), where it's trying to get the PID bytes of the first and last packets in the transaction.

martinling:
I found the bug!

The problem was with the way I was implementing the `target_range` method for `CompactSnapshot`.

I was forwarding this to the `target_range` method on `CompactReader` and passing in the correct `target_length` parameter for the length of the target stream in the snapshot.

However, the `target_range` implementation on `CompactReader` ignores that parameter unless it thinks it needs it, so the resulting range wasn't being restricted to the length of the target stream in the snapshot.

The fix is in commit a0712bfa92b98fbfdeb7def11ba4f635a0158aa5.

I've kept the change to error handling on item updates, but removed the subsequent debug code.

This should now be ready to go.

miek (approved):
Looks good to me, thank you!

PR #239, opened 2025-07-05
Reimplement main window as a subclass with UI loaded from file

### LINKED ISSUE #132: Window too small

martinling:
Up till now, the main window has been constructed programatically in the `activate` function.

This PR moves that code into a new `PacketryWindow` GObject class which subclasses GTK's `ApplicationWindow`.

The motivation here is that we want to set the initial positions of the horizontal and vertical pane separators to a sensible proportion of the initial window size. Currently we fudge this with some hardcoded minimum sizes in pixels. This has caused various frustrations discussed in #191.

By subclassing `ApplicationWindow`, we can override the `size_allocate` method. This seems to be the only way in GTK4 to consistently be notified when the window size is first known.

With that method hooked, we can set the initial size of the traffic pane to 75% of the initial window size both horizontally and vertically.

Additionally, this PR integrates and extends the work from #231 to define the main window layout in an XML UI file that can be edited visually with the [Cambalache](https://blogs.gnome.org/xjuan/category/programming/cambalache-ui-maker/) tool.

Fixes #132, supersedes #191 and #231.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -32,9 +32,12 @@ mod usb;
 mod util;
 mod version;
 
-use gtk::prelude::*;
+// The GTK version we require at runtime.
+const GTK_REQUIRED: (u32, u32, u32) = (4,12,0);
+
+use gtk::{prelude::*, ApplicationWindow, Label, Button, Orientation};
 use gtk::gio::ApplicationFlags;
-use gtk::glib::{self, OptionArg, OptionFlags};
+use gtk::glib::{self, clone, OptionArg, OptionFlags};
 
 use testing::test_cynthion;
 use ui::{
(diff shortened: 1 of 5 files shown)


martinling:
Crashes on Windows, marking as draft for now.

martinling:
The Windows crash was due to my test system still having GTK 4.10.5 DLLs in the `PATH`, when the UI file specifies that GTK 4.12 is required.

This resulted in a cryptic crash because of:
1. Lack of console output when running the `packetry.exe` GUI binary; using `packetry-cli.exe` to attach a console shows a panic and backtrace.
2. The panic message from the gtk-rs template macro being misleading in this situation. I've opened an issue for this:
https://github.com/gtk-rs/gtk4-rs/issues/2095

Fixed in https://github.com/greatscottgadgets/packetry/pull/239/commits/540f6dc62f04f3b6d01a8f3af8d7e3c437a07670 by checking the runtime GTK version at startup, and displaying an crudely thrown together error window instead of crashing, if the version requirement is not met.

image

As such this is now ready for review and merge again.

antoinevg (approved):
Code looks great. Cambalache looks quite cool!

Tested on Windows 10, Mac, Linux with:

* Release gateware.
* Pre-release gateware with power control support.

PR #255, opened 2025-08-27
Add filename extensions automatically

### LINKED ISSUE #176: Automatically add pcap file extension when saving files
Since the file format for logs is pcap, it makes sense that a ".pcap" file extension should be automatically added to the end of the filename when saving the file. Likewise when loading log files it shoud look for *.pcap files by default.

martinling:
Fixes #176.

When saving a capture, if no filename extension is entered, the PcapNg format will be selected and `.pcapng` will be added to the filename.

When saving payload or descriptor data, if no filename extension is entered, `.bin` will be added to the filename.

### DIFF
--- a/src/ui/mod.rs
+++ b/src/ui/mod.rs
@@ -642,6 +642,31 @@ fn choose_file<F>(
         chooser.connect_response(move |dialog, response| {
             if response == gtk::ResponseType::Accept {
                 if let Some(file) = dialog.file() {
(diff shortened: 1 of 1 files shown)


miek:
One slight nitpick I have on this is that it, if you try and save to an existing filename, it bypasses the usual "are you sure?" warning that you'd get from the file picker. I'm not sure if there's a good solution to that though?

miek (approved):

Issue #262, opened 2025-09-03
Crash during low speed capture


miek:
If I set up the same situation as in https://github.com/greatscottgadgets/packetry/pull/227#issuecomment-3244923590, except with a low-speed keyboard plugged in, and I capture/stop several times eventually it crashes (maybe 1 in 10 times) with:

```
attempt to subtract with overflow
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panicking.rs:75:14
   2: core::panicking::panic_const::panic_const_sub_overflow
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panicking.rs:175:17
   3: <packetry::util::id::Id<T> as core::ops::arith::Sub>::sub
             at ./src/util/id.rs:85:7
   4: packetry::database::compact_index::CompactWriter<Position,Value,_>::push
             at ./src/database/compact_index.rs:236:29
   5: packetry::decoder::Decoder::group_append
             at ./src/decoder.rs:1100:21
   6: packetry::decoder::Decoder::group_update
             at ./src/decoder.rs:1026:17
   7: packetry::decoder::Decoder::transaction_end
             at ./src/decoder.rs:872:17
   8: packetry::decoder::Decoder::transaction_update
             at ./src/decoder.rs:782:17
   9: packetry::decoder::Decoder::handle_raw_packet
             at ./src/decoder.rs:587:9
  10: packetry::ui::start_capture::{{closure}}::{{closure}}
             at ./src/ui/mod.rs:958:25
  11: packetry::ui::start_capture::{{closure}}::{{closure}}
             at ./src/ui/mod.rs:979:27
```

I patched the subtraction as described there and saved out the pcapng:

[newcrash.zip](https://github.com/user-attachments/files/22124110/newcrash.zip)

martinling:
Oof 😢. OK, I'll take a look.

PR #227, opened 2025-03-05
Non-packet events from Cynthion USB analyzer

martinling:
This PR adds support for non-packet events that may be captured by Cynthion or other USB packet analyzers.

The scope here is limited to events related to the USB capture process itself, or that occur on the USB bus. This implies that these events have a clearly defined position within the stream of captured USB packets.

Events that occur independently of the USB bus, such as logic signals changing on the Cynthion PMOD ports, or voltage/current data points captured from the PAC1954 chip, are out of scope for this event mechanism because they can e.g. occur mid-packet, and therefore require independent timestamping and indexing. 

The proposed set of initially supported events is:
```rust
CaptureStop(Requested) => "Capture stopped by request",
CaptureStop(BufferFull) => "Capture stopped due to full buffer",
CaptureStop(Error) => "Capture stopped due to an error",
CaptureStart(High) => "Capture started at High Speed (480 Mbps)",
CaptureStart(Full) => "Capture started at Full Speed (12 Mbps)",
CaptureStart(Low) => "Capture started at Low Speed (1.5 Mbps)",
CaptureStart(Auto) => "Capture started with automatic speed selection",
SpeedChange(High) => "Speed changed to High Speed (480 Mbps)",
SpeedChange(Full) => "Speed changed to Full Speed (12 Mbps)",
SpeedChange(Low) => "Speed changed to Low Speed (1.5 Mbps)",
SpeedChange(Auto) => "Speed changed to automatic selection",
LineStateChange(SE0)  => "SE0 line state detected",
LineStateChange(ChirpJ) => "Chirp J state detected",
LineStateChange(ChirpK) => "Chirp K state detected",
LineStateChange(ChirpSE1) => "Chirp SE1 state detected",
LineStateChange(LsJ) => "Low Speed idle state detected",
LineStateChange(LsK) => "Low Speed resume state detected",
LineStateChange(FsJ) => "Full Speed idle state detected",
LineStateChange(FsK) => "Full Speed resume state detected",
LineStateChange(SE1) => "Invalid SE1 line state detected",
VbusInvalid => "VBUS voltage became invalid",
VbusValid   => "VBUS voltage became valid",
LsAttach => "Device attached at Low Speed",
FsAttach => "Device attached at Full Speed",
BusReset => "Bus reset",
DeviceChirpValid => "Device HS chirp validated",
HostChirpValid => "Host HS chirp validated",
Suspend  => "Bus entered suspend",
Resume   => "Resume signal detected",
[...]

PR #263, opened 2025-09-08
Catch panics in decoder thread and display a dialog, rather than crash

martinling:
This change makes issues like bug #262 and https://github.com/greatscottgadgets/packetry/pull/227#issuecomment-3244923590 less disruptive.

When a helper thread spawned from the UI panics, the panic now be caught as an error via `catch_unwind`. The operation will be stopped cleanly and an error message displayed. The application will not crash and the UI will remain usable.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -1,6 +1,7 @@
 //! The backend API for USB capture devices.
 
 use std::collections::BTreeMap;
+use std::panic::UnwindSafe;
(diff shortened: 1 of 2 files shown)


miek (approved):
Looks good!

One extra change that might be useful is to put the panic message in the text view too so it can be copy&pasted to an issue along with the backtrace, but I'm happy with merging this as-is and discuss/change that later.

Issue #206, opened 2024-10-28
nit: reduce `ListItem` spacing so tree branches visually connect


schnommus:
This is a very minor look-and-feel thing. Looking for opinions, happy to help out (e.g. PR) if you think it would be useful.

A normal packetry session might look like this:
![Screenshot_select-area_20241028221444](https://github.com/user-attachments/assets/779b71e2-da34-46f3-be71-dbfbbbb11083)

By switching to a bitmap monospace font (`Terminus` here) and reducing the vertical padding, the tree connections between rows meet up:
![Screenshot_select-area_20241028221329](https://github.com/user-attachments/assets/bd635fb7-d89c-443c-a719-a22145a093e8)

There are still some horizontal tree misalignments above (which would need to be fixed). It would also make sense to only reduce padding in the central `ListView` and not everywhere in the whole application of course (this is happening as I'm applying the style to every element using the GTK debugger, however is just for the sake of example)

Does this sound like a reasonable change or are there some plan to move to different connector icons? (i.e graphical, non-ASCII)

martinling:
Hi @schnommus,

Yes, it would be nice if we could get those to join up nicely.

In the longer term, I'd been thinking we'd substitute each of those text characters with graphical elements, so that we have more control over their appearance. But the rendering is always going to have to work on a per-row basis, so we'll need to eliminate the padding between rows either way.

The easiest way to do it is probably to eliminate all the vertical padding which gets between each rows' label containing the connecting lines, and then increase the font size of that label such that it ends up defining the overall height of the row, whilst leaving the rest of the content at the same size and spacing as it is now.

Even if we eventually implement a graphical version, I'd want to keep the text version behind the scenes. It's good for outputting the UI contents to a file for test purposes, and we might want a TUI frontend at some point in the future too.

martinling:
By the way - I notice you moved the expander widgets to the left of the connecting lines too.

We used to have it that way, but we changed it to keep the expanders on the right, because in a complex capture the connections can get quite complicated, which would put the expanders a long way from the things they act on:

![image](https://github.com/user-attachments/assets/041f1106-742d-410d-a958-2788cc11cf53)

schnommus:
Hi @martinling,

Indeed as you suggested increasing the font size *only* in the connector cell makes this easier to achieve. Unfortunately I was not able to get the padding to exactly zero by modifying the various component properties, and had to apply a CSS rule to the label instead with a negative margin to get truly 'no' padding.

[...]

PR #273, opened 2025-11-12
ui: Add a widget to draw each item connectors

jmaselbas:
~~this is a work-in-progress:
The connectors are always drawn in black, even with a dark theme, as it doesn't follow the widget's color...~~

edit: the connectors color now follows the foreground color

Related to #206

### DIFF
--- a/src/ui/item_widget.rs
+++ b/src/ui/item_widget.rs
@@ -60,6 +60,8 @@ impl ItemWidget {
         wrapper.add_controller(right_click);
         wrapper.imp().text_label.replace(
             Label::builder()
+                .margin_top(2)
+                .margin_bottom(2)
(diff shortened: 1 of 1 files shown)


jmaselbas:
it looks like this:
pak

martinling:
Removing the padding between cells causes the text to get quite tightly packed vertically, particularly on macOS:
Image 07-12-2025 at 17 37
Note also the graphics for the "End of control transfer" item are offset.

To regain the lost spacing, I've pushed a commit to this branch which sets `margin-top` and `margin-bottom` of the text labels to 2px:
Image 07-12-2025 at 17 36

On my Linux system this now appears as:
image

And on Windows:
margin-2

It would be nice to have things a bit more consistent across systems but I'm not sure what the best way to achieve that is.

PR #277, opened 2025-12-07
CI: update vcpkg to fix pango build failure.

martinling:
Our Windows builds have been failing due to https://github.com/microsoft/vcpkg/issues/47984.

This should fix it by updating to include the fix from https://github.com/microsoft/vcpkg/pull/48454.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -61,7 +61,7 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     env:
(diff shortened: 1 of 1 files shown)


martinling:
Same problem on `gdk-pixbuf`, so we need https://github.com/microsoft/vcpkg/pull/48679 too, and probably https://github.com/microsoft/vcpkg/pull/48711 too, but that's still in draft.

PR #286, opened 2026-04-09
ui: Set expander to be vertically centered

jmaselbas:
align-fix


martinling:
Ah right, that must have ended up a little bit off after #273.

Well spotted, thank you!

jmaselbas:
i think it was already the case before but the expander visuals recently changed and it is more noticable

PR #293, opened 2026-07-17
quit on cmd+q on macos.

carlossless:
Registers the app.quit action so the native macOS menu's Quit item (Cmd+Q) works.


martinling (approved):
Thank you!

PR #294, opened 2026-07-17
constrain sizing of the traffic and device panes.

carlossless:
Prevents either pane from collapsing and gives extra width to the traffic pane on resize.

### DIFF
--- a/src/ui/packetry.ui
+++ b/src/ui/packetry.ui
@@ -167,6 +167,7 @@ CONTROL</property>
                   <object class="GtkBox">
                     <property name="orientation">vertical</property>
                     <property name="vexpand">True</property>
+                    <property name="width-request">320</property>
                     <child>
                       <object class="GtkStackSwitcher" id="switcher">
                         <property name="hexpand">True</property>
@@ -218,11 +219,16 @@ CONTROL</property>
                 </child>
                 <child>
                   <object class="GtkScrolledWindow" id="device_window">
+                    <property name="width-request">200</property>
                     <property name="child">
                       <object class="GtkListView"/>
                     </property>
                   </object>
                 </child>
+                <property name="shrink-start-child">False</property>
+                <property name="shrink-end-child">False</property>
+                <property name="resize-start-child">True</property>
+                <property name="resize-end-child">False</property>
               </object>
             </child>
             <child>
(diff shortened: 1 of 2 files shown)


martinling:
I tried this, and it's mostly an improvement over the behaviour we had before, but:
1. It would be nice to have the same behaviour for the vertical split as well as the horizontal one; I added a commit which applies the same property changes to achieve this.
2. Resizing below a certain width triggers a lot of warnings to the console of the form:
`Trying to measure GtkPaned 0x56173bed6460 for width of 543, but it needs at least 605`
It would be nice to stop that happening.

Ideally, I would like to find a solution that makes the resize behaviour reversible: i.e. if you shrink the window from size A to size B < A, then enlarge it back to A, everything should go back to the same positions. This is particularly relevant when using a tiling window manager. I had a go at making this the case, by modifying the `size_allocate` method on `PacketryWindow` to try to maintain pane ratios, but wasn't able to get it to work the way I wanted yet.

carlossless:
@martinling I took a shot at addressing that. Let me know what you think?

martinling:
@carlossless Ooh yeah, this feels just right. Nicely done! Any thoughts on the warning messages?

carlossless:
@martinling I am actually not able to reproduce those warnings. I'm using hyprland myself (also a tiling wm) and stuffing the window into a smaller space than its intrinsic size just starts clipping the sides of the window - no warnings.

What are you using? Any env vars should be set other than the default debug ones?

martinling:
Hmm, interesting. Personally I'm using [XMonad](https://xmonad.org/) out of longstanding habit. I see the warnings both when the window shrinks via tiling, and when grabbing it out of the tiling layout to resize arbitrarily, so I would have thought this was something WM-independent.

carlossless:
I looked into this further, and it seems to be a matter of the window manager not respecting WM_NORMAL_HINTS/PMinSize on X11. I tested this out myself, and I saw XMonad ignore it, crush the interface (notice the top bar in the screenshot) and print those warnings.

image

I'm inclined to think that this is more of an issue with the window manager, rather than packetry.

martinling:
Makes sense, I've just tried this branch with the default window manager on Ubuntu and it resists being shrunk below the minimum there.

Let's merge this then - thank you!

PR #297, opened 2026-07-28
Fix compact index iteration with an incomplete final segment

### LINKED ISSUE #290: Failed to load segment width with "stream length = request length -1"
Edit: Just realized this issue actually was mentioned before (https://github.com/greatscottgadgets/packetry/issues/213) and closed. But here I am now hitting this bug quite a lot.

Image

<br>

### Software: 
Packetry 0.5.0 Linux AppImage

<br>

### Issue:

I often get this error message when trying to export the recorded data to a pcap file.

The error is always caused by "`stream length` = `request length` -1", like in the screenshot `stream length` = 3433 but `request length` = 3434.
Packetry will refuse to process the data and result in export data only few bytes large. Nothing else can be done to save the captured data.

This usually happens when I "stopped the recording safely" by pressing the stop button (and it actually stopped.)

Interesting though, if I "force stop" the recording by disconnecting usb connection between cynthion and my PC, this issue rarely happens. (Sometimes I have to do this because Packetry will simply freeze after I clicked the stop button.)

showier-drastic:
A newly started compact-index segment has a base value but no delta width until its second value is written. Avoid requesting a nonexistent width entry when iterating across such a final segment.

Add a regression test covering 18 segments backed by only 17 width entries.

This is reproduced on a problematic USB device that keeps resetting.

### DIFF
--- a/src/database/compact_index.rs
+++ b/src/database/compact_index.rs
@@ -434,7 +434,15 @@ where
         let segment_starts = self.segment_start_reader.iter(&seg_range);
         let base_values = self.segment_base_reader.iter(&seg_range);
         let data_offsets = self.segment_offset_reader.iter(&seg_range);
-        let segment_widths = self.segment_width_reader.iter(&seg_range);
+        // A segment only gets a width when its first delta is written. The
+        // final committed segment can therefore have a base value but no
+        // corresponding width entry yet.
+        let width_end = min(
+            seg_range.end,
+            SegmentId::from(self.segment_width_reader.len())
+        );
+        let width_range = seg_range.start..width_end;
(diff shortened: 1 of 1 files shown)


martinling:
Thank you so much! This had been causing weird crashes for a long time.

A failure when trying to load one too many entries from the `segment_widths` stream had been reported twice, first in #213 where it was unreproducible, and again in #290 - from which we got database dumps demonstrating the problem, but I still didn't spot the underlying bug.

I've now verified that with the change in this PR, I can restore a capture from the database dumps that @DrmnSamoLiu sent me in #290.

martinling (approved):

PR #299, opened 2026-08-02
Upgrade dark-light to 2.0.0 for correct XDG portal theme detection

carlossless:
Upgrades `dark-light` 1.1.1 -> 2.0.0 so the preferred colour scheme is detected on desktops that report it only via the XDG desktop portal (Hyprland, Sway, Xfce, and other environments not special-cased by dark-light).

`dark-light` 1.1.1 mis-parses the portal's `Settings.Read` reply, which returns the value inside two layers of variant, and falls back to reading GNOME/KDE-specific settings. On desktops it doesn't recognise, `detect()` returns `Default` and Packetry always starts light. 2.0.0 reads the portal via `ashpd`, which handles the nested variant correctly.

I was experiencing this issue myself on Hyprland.


martinling (approved):
Looks good to me.

